PR #156, opened 2015-02-12
Update licence to correct format and date

tombye:
As per instructions from @annashipman, the copyright for this project should read "Crown Copyright", not "HM Government".


LICENCE
```diff
@@ -1,4 +1,4 @@
-Copyright (c) 2014 HM Government
+Copyright (c) 2015 Crown Copyright (Government Digital Service)
```
bradleywright: Should be `2012`, as that's when this repo was created.

PR #157, opened 2015-02-19
Mobile-first typography

gemmaleigh:
- Set `$font-size-640` and `$line-height-640` first
- Override with `$font-size` and `$line-height` for larger screens
- Do `$is-print` check last, to override any font-size set with `$font-size-print`

### DIFF
--- a/stylesheets/_typography.scss
+++ b/stylesheets/_typography.scss
@@ -27,11 +27,11 @@ $is-print: false !default;
   }
   font-size: $font-size-640;
   line-height: $line-height-640;
+  font-weight: $font-weight;
   text-transform: none;
   @media (min-width: 641px) {
     font-size: $font-size;
     line-height: $line-height;
-    font-weight: $font-weight;
   }
   @if $is-print {
     font-size: $font-size-print;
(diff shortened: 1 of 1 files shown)


edds:
The problem with doing that is font sizes in older versions of Internet Explorer which don't understand media queries will be getting small font-sizes.

gemmaleigh:
I thought that by setting the $is-ie: true; flag in each IE_8/7.6/.scss file, when using the @media mixin from the _conditionals.scss file, then IE will get all the styles declared in the media block - so those declared last - the larger sizes will override the smaller sizes.

edds:
This isn't using the media mixin due to limitations in Sass. This is using a pure `@media` approach.

edds:
You could possibly get around it by using the if IE stuff for this. If you wanted to still print out the bigger sizes for IE.

gemmaleigh:
Sorry @edds my mistake. I'll amend it to do the `$is-ie` check here.

alicebartlett:
[comment from the frontend meeting] @edds are you happy to merge this now or do you have further comments?

PR #158, opened 2015-02-23
Allow a transform-origin param to be passed to scale mixin

jonnywyatt:
Defaults to using the center as the origin.

### DIFF
--- a/stylesheets/_css3.scss
+++ b/stylesheets/_css3.scss
@@ -20,13 +20,17 @@
           box-shadow: $shadow;
 }
 
-@mixin scale($x, $y) {
+@mixin scale($x, $y, $transform-origin: 50% 50% 0) {
   // $x and $y should be numeric values without units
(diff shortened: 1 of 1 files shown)


stylesheets/_css3.scss
```diff
@@ -20,13 +20,19 @@
...
   -webkit-transform: scale($x, $y); // Still in use now, started at: Chrome 4.0, Safari 3.1, Mobile Safari 3.2, Android 2.1
      -moz-transform: scale($x, $y); // Firefox 3.5 to 15.0
       -ms-transform: scale($x, $y); // IE9 only
        -o-transform: scale($x, $y); // Opera 10.5 to 12.0
           transform: scale($x, $y);
+
+  -webkit-transform-origin: $transform-origin;
+     -moz-transform-origin: $transform-origin;
+      -ms-transform-origin: $transform-origin;
+       -o-transform-origin: $transform-origin;
```
edds: It would be good to get comments on these vendor prefix versions like above detailing which versions of those browsers need the prefix. Also I am not sure we need the opera one where they added a non-prefixed version in 12.10 and checking our stats we don't actually get any users of opera that old.

PR #159, opened 2015-02-24
Don’t print phase banners

alextea:
The phase banner (alpha/beta banner) should not be printed as it is not relevant to the printed page.

### DIFF
--- a/stylesheets/design-patterns/_alpha-beta.scss
+++ b/stylesheets/design-patterns/_alpha-beta.scss
@@ -36,6 +36,10 @@
       vertical-align: baseline;
     }
   }
+
+  @if $is-print == true {
+    display: none;
+  }
 }
 
 // Phase tag usage:


edds:
Is it always not relevant? We use the phase banners to indicate that content is still being transition and that it isn't the canonical version of that content ([like on the current AAIB reports](https://www.gov.uk/aaib-reports/aaib-investigation-to-gulfstream-550-hz-a6)). Having that printed to tell the user that the report isn't the main copy of the report would be useful I would have thought.

alextea:
I don't know, I opened this PR to start a conversation. In the situation that triggered this, it doesn't seem to add much as there is no alternative service.

edds:
I think this should be done on a case by case basis rather than globally as we can't be sure it won't always be helpful. If anyone has any other thoughts feel free to reopen it.

PR #160, opened 2015-03-09
Analytics

jonnywyatt:
Moved the new analytics library from static so it can be easily used in non-rails projects.
[PR for analytics library](https://github.com/alphagov/static/pull/549)
The tests for analytics used Jasmine 2 so it made sense to upgrade the existing frontend_toolkit tests to use that.
Analytics had one usage of jQuery so replaced that with plain Javascript so it has no dependencies.
Included a sample init to be copied into your own project as GA account IDs should not be stored in this repo.


javascripts/govuk/analytics/tracker.js
```diff
@@ -0,0 +1,81 @@
+(function() {
+  "use strict";
+  window.GOVUK = window.GOVUK || {};
+
+  var Tracker = function(config) {
+    var setDimension = this.setDimension.bind(this);
```
fofr: In static `.bind` is using a polyfill for older browsers, this might not work unless the polyfill is present.

javascripts/govuk/analytics/sample-init.js
```diff
@@ -0,0 +1,20 @@
+(function() {
```
easternbloc: I think this should live in a folder called `/examples` so it's very clear what this is doing

javascripts/govuk/analytics/tracker.js
jonnywyatt: Good point.. maybe I'll use jQuery proxy instead to avoid bringing more code over?

easternbloc:
Other than the comments already made LGTM

javascripts/govuk/analytics/tracker.js
```diff
@@ -0,0 +1,81 @@
...
+
+    function setPixelDensityDimension() {
+      var pixelRatioDimensionIndex = 11;
+
+      if (window.devicePixelRatio) {
+        setDimension(pixelRatioDimensionIndex, window.devicePixelRatio, 'Pixel Ratio', 2);
+      }
+    }
+
+    function setHTTPStatusCodeDimension() {
```
fofr: I'm wondering whether `setPixelDensityDimension` and `setHTTPStatusCodeDimension` can be omitted here, and setup using the new `prePageviewConfiguration` by apps that need it. It would make the tracker more generic and avoid the need for `.bind` on `setDimension`.

easternbloc: +1 for that. The bind is slightly confusing what it's doing initially.

javascripts/govuk/analytics/tracker.js
```diff
@@ -0,0 +1,81 @@
...
+    if (typeof index !== "number") {
+      index = parseInt(index, 10);
+    }
+
+    if (typeof scope !== "number") {
+      scope = parseInt(scope, 10);
+    }
+
+    this.universal.setDimension(index, value);
+    this.classic.setCustomVariable(index, value, name, scope);
```
fofr: There's a presumption here that is worth highlighting. When setting custom dimensions (custom variables in classic analytics), it assumes that the universal and classic variables/dimensions are setup with the same index. This is true for the GOV.UK profiles that are being tracked, but not necessarily for other apps that could use this code.

eg the Section dimension is at index 1, and the Section classic custom variable is at slot 1.

jonnywyatt: I'll remove these until there's more certainty around this

edds:
Couple of thoughts:

This needs some better documentation as to what it is and how to use it. I would suggest creating a `docs` folder in the project root and creating some documentation as to what the different parts are and how you can use them.

I know it has now been removed but relying on `bind` in things that are in the toolkit is fine. The toolkit [contains the bind polyfill](https://github.com/alphagov/govuk_frontend_toolkit/blob/master/javascripts/vendor/polyfills/bind.js) for people that need to use it. This should just be made clear in the documentation for that feature that you will need to include it.

javascripts/govuk/analytics/tracker.js
edds: Can we just add this to the documentation rather than removing this. If someone else wants to use this and has different requirements we can have another look at how this is implemented but for the moment this will work for all the people that want this code.

javascripts/govuk/analytics/print-intent.js
```diff
@@ -0,0 +1,31 @@
...
+    "use strict";
+    var printAttempt = (function() {
+        GOVUK.analytics.trackEvent('Print Intent', document.location.pathname);
+    });
+
+    // Most browsers
+    if (window.matchMedia) {
+        var mediaQueryList = window.matchMedia('print'),
+            mqlListenerCount = 0;
+        mediaQueryList.addListener(function(mql) {
```
edds: the whitespace is a little out here. The convention for this project is two space indents.

javascripts/govuk/analytics/tracker.js
jonnywyatt: setDimension was restored. the config.prePageviewConfiguration callback is available for the end user to call it before the page view is tracked.

javascripts/govuk/analytics/print-intent.js
fofr: @edds Looks to be the same in static where it was ported from.

fofr:
@edds I've tidied up some of the docs in https://github.com/alphagov/govuk_frontend_toolkit/pull/161, I'll add analytics details when that PR is merged.

fofr:
Closing while I write documentation.

PR #161, opened 2015-03-09
Split mixin and javascript documentation into separate files

fofr:
Add a `/docs` folder for documentation, and reduce the clutter within the main `README` by moving information about mixins and javascript into separate files.

screen shot 2015-03-09 at 14 00 49


edds:
I think that it might be worth splitting them down into even more files (specifically the JavaScript ones into a file per feature). But this is a vast improvement on what is there already :+1:

PR #162, opened 2015-03-10
Port analytics API from static

fofr:
Moved the new analytics library, javascript error tracking and print intent tracking from static so it can be easily used in non-rails projects.

Original PR for analytics library: https://github.com/alphagov/static/pull/549
Earlier closed PR for context: https://github.com/alphagov/govuk_frontend_toolkit/pull/160

Changes from the static version:
* dependencies on jQuery removed
* dependency on the `.bind` polyfill removed
* Tracker made more generic
  * GOVUK specific custom dimensions removed
  * Pageview tracking happens within the `init` so that apps can set custom dimensions beforehand

Also included:
* The tests for analytics used Jasmine 2 so it made sense to upgrade the existing frontend_toolkit tests to use that.
* Included a sample init to be copied into your own project as GA account IDs should not be stored in this repo.
* Documentation


javascripts/govuk/analytics/example-init/example-init.js
```diff
@@ -0,0 +1,24 @@
...
+  // Configure profiles, setup custom vars, track initial pageview
+  // and make interface public for virtual pageviews and events
+  GOVUK.analytics = new GOVUK.Tracker({
+    universalId: 'UA-XXXXXXXX-X',
+    classicId: 'UA-XXXXXXXX-X',
+    cookieDomain: cookieDomain
+  });
+
+  // Set custom dimensions before tracking pageviews
+  // GOVUK.analytics.setDimension(…)
```
jonnywyatt: Is it worth adding to the comment, the URL to the doc section about using setDimension? Just so they know which params to pass to it.

jonnywyatt:
:+1: Really good documentation

edds:
I would remove the `javascripts/govuk/analytics/example-init/example-init.js`. It is a duplicate of what is in the docs and has the potential to diverge from the docs. Having it held within the docs will reduce the potential for this divergence.

Can you add some documentation around the two files which aren't mentioned (print tracking and error reporting) what they do and how they should be used.

The formatting of the list in the "Create an analytics tracker" section of the docs is slightly broken. I think you are missing a blank line.

jonnywyatt:
Agreed the example is duplicated but if you were going to keep one copy wouldn't it be the file, for convenience and clarity of copying it out? The annotations from the readme could move into it as code comments?

edds:
I don't think that the example should live with the code in the `javascripts/govuk/analytics` folder. If it wants to live as a separate file it should live in the `docs` folder. Having example code next to real code is is a mix of concerns that shouldn't really exist. Personally I think it works best in the markdown document file so there is a single place which explains how to do everything. There is already code in the docs you would want to copy paste out if you were setting this up in a new project.

jonnywyatt:
Yep, agreed it should go under docs. I'd say a separate file is clearer but up to you guys :)

fofr:
Updated in bff50a4 to have a single canonical init example within the docs, also linked to from the source file.

tombye:
This is a great pull request, really detailed but easy to read and understand :+1: 

@jonnywyatt There is a mention of the lack of dependencies in the analytics page of the docs: https://github.com/alphagov/govuk_frontend_toolkit/pull/162/files#diff-a18afd2d1d40d1e6a39ba0d5f8fc18ddR21.

dsingleton:
:+1:

fofr:
Is this good to merge?

PR #164, opened 2015-03-11
Analytics - error tracking and print intent - don't run on load

jonnywyatt:
As the JS for these two analytics plugins will often be bundled by default, they shouldn't run automatically. This is because they might not have been loaded after the main Tracker file.
So instead, they should be run explicitly from the code in the analytics 'init' file, after the Tracker has been set up. See [the docs](https://github.com/alphagov/govuk_frontend_toolkit/blob/master/docs/analytics.md) for details and example.


fofr:
Looks good to me. (Diff best viewed with `?w=1` in the URL)

@edds do you have any thoughts?

PR #168, opened 2015-03-16
Version bump to 3.3.0

fofr:
* Add: Analytics - pageview tracking for a print attempt


VERSION.txt
```diff
@@ -1 +1 @@
-3.2.1
+3.3.1
```
benilovj: This should be `3.3.0`.

PR #169, opened 2015-04-07
Update Multivariate library to work with the GOVUK Analytics API

binaryberry:
Pivotal Tracker: 
https://www.pivotaltracker.com/story/show/91414734

### DIFF
--- a/javascripts/govuk/multivariate-test.js
+++ b/javascripts/govuk/multivariate-test.js
@@ -71,18 +71,16 @@
   };
 
   MultivariateTest.prototype.setCustomVar = function(cohort) {
-    window._gaq = window._gaq || [];
-    window._gaq.push([
-      '_setCustomVar',
-      this.customVarIndex,
-      this.cookieName(),
-      cohort,
+    GOVUK.analytics.setDimension(
+      this.customVarIndex, 
+      this.cookieName(), 
+      cohort, 
       2 // session level
-    ]);
+    ); 
     // Fire off a dummy event to set the custom var on the page.
     // Ideally we'd be able to call setCustomVar before trackPageview,
     // but would need reordering the existing GA code.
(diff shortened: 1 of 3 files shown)


binaryberry:
cc @fofr

javascripts/govuk/multivariate-test.js
```diff
@@ -71,18 +71,11 @@
...
   MultivariateTest.prototype.setCustomVar = function(cohort) {
-    window._gaq = window._gaq || [];
-    window._gaq.push([
-      '_setCustomVar',
-      this.customVarIndex,
-      this.cookieName(),
-      cohort,
-      2 // session level
-    ]);
+    GOVUK.analytics.setDimension(this.customVarIndex, this.cookieName(), cohort, 2) // session level
```
edds: Can this be split out on new lines like it was previously so that it is obvious that the `2` is the part that the `// session level` comment relates to.

binaryberry: Thanks edds! We fixed it now.

javascripts/govuk/multivariate-test.js
```diff
@@ -71,18 +71,16 @@
...
+    GOVUK.analytics.setDimension(
+      this.customVarIndex, 
+      this.cookieName(), 
+      cohort, 
       2 // session level
-    ]);
+    ); 
     // Fire off a dummy event to set the custom var on the page.
     // Ideally we'd be able to call setCustomVar before trackPageview,
     // but would need reordering the existing GA code.
```
fofr: Does this comment still apply? The new analytics work means custom variables _can_ be set before the initial pageview is tracked.

edds: There is no guarantee that it will be run before the initial pageview is tracked. In an ideal world I think ensuring the dimensions are pushed to Google could be abstracted away inside our `GOVUK.analytics` object so we don't need to worry about it in other modules.

PR #176, opened 2015-04-14
Change error colour to a darker red

gemmaleigh:
Change $error-colour from $mellow-red: #df3034 to the darker #af1324.

On a white background, #af1324 has a contrast ratio of 7:14:1 and
meets level AAA, this should provide enough contrast between text and
its background so that it can be read by people with moderately low
vision.

http://www.w3.org/WAI/WCAG20/quickref/#qr-visual-audio-contrast7

This will be used for error text and also for the border to the left of
an error, or the border of an error summary box.


gemmaleigh:
Hi @edds, this was removed from master - can it please go back in?

tombye:
This is a re-issue of this pull request: https://github.com/alphagov/govuk_frontend_toolkit/pull/171, the commits for which got removed from master.

PR #183, opened 2015-04-16
Bump version to 3.4.2 - multivariate test bug fix

binaryberry:
Updating gem version following bug fix. Before this fix, when a user fell into variant 0 of a multivariate test, the data wouldn't be reported to Google correctly because of a broken null check in the code block that opts the user into the Google Content Experiment.

cc @benilovj


govuk_frontend_toolkit
```diff
@@ -0,0 +1 @@
+Subproject commit 6e41f736e04e3416654f726f068c30f8a3cc82d4
```
edds: I don't think you meant to add this. I am going to guess you have a copy of the frontend toolkit checked out within your frontend toolkit folder.

binaryberry: Thanks Edds! A remnant from my git troubles. All clear now.

PR #185, opened 2015-05-01
Add cross domain tracking to Tracker object

bradleywright:
So that pages that link to known partners can allow those partners to track full user journeys, this commit adds the necessary GA and UA incantations to link domains and separate accounts together.

This is to enable the removal of bespoke HTML code in several transaction start pages across GOV.UK.

One thing I'm not sure about - do we need to support separate UA and GA account IDs? @fofr @benilovj

### DIFF
--- a/docs/analytics.md
+++ b/docs/analytics.md
@@ -160,3 +160,15 @@ Pull `print-intent.js` into your project, and initialise it after analytics (see
 ## Error tracking
 
 Pull `error-tracking.js` into your project, and initialise it after analytics (see [Create an analytics tracker, above](#create-an-analytics-tracker)), to track JavaScript errors.
+
+## Tracking across domains
+
+Once a Tracker instance has been created, tracking across domains can be set up
+for pages like:
+
+```js
+GOVUK.analytics.addLinkedTrackerDomain(trackerIdHere, nameForTracker, domainToLinkTo);
+```
+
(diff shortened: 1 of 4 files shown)


edds:
This looks good to me but would appreciate @fofr having a look as he has done the bulk of the GA work.

tombye:
Looks great. Would it be possible to add something to the [analytics docs](https://github.com/alphagov/govuk_frontend_toolkit/blob/master/docs/analytics.md) for this?

bradleywright:
@tombye added in 262b168

tombye:
@bradleywright thanks!

fofr:
As far as I know the cross-domain trackers have a single Universal profile (eg https://github.com/alphagov/frontend/blob/master/app/views/root/_transaction_cross_domain_analytics.html.erb). Which means we won't need the classic parts,  `this.classic.addLinkedTrackerDomain(trackerId, name, domain);`.

Creating a classic and universal tracker for the same profile may have unexpected results. It looks like this code would track the same pageview twice – once through the classic mechanism and once through the Universal one. For a Universal analytics profile, I don't know how Google handles data passed to it using the classic API.

My suggestion is to remove the classic implementation.

bradleywright:
@fofr updated.

fofr:
:+1:

PR #186, opened 2015-05-05
Add sass variables for discovery and live phases

edwardhorsford:
This PR adds a named SASS variable for the discovery and live phases of development, in the same format as the existing $alpha-colour and $beta-colour variables.

We don't often use these colours, but for consistency it would be good for them to be included.


alexmuller:
Makes sense to me :+1:

PR #189, opened 2015-05-07
Make tests run in browser

alicebartlett:
The upgrade to `grunt-contrib-jasmine 0.8.2` in https://github.com/alphagov/govuk_frontend_toolkit/commit/5fa6d6d522fa9a4996c3ffd50c4d6808f6b0615a updated Jasmine to 2.0.0 which broke the local test runner.

This PR fixes the test runner to work with Jasmine 2.0.0


tombye:
It's a bit pedantic to mention but the original bump for `grunt-contrib-jasmine` brought the version of Jasmine up to `2.0.1`.

This has no bearing on the correctness of this pull request which is great work :+1:

PR #190, opened 2015-05-07
Make tracker.js more liberal in what it accepts

alicebartlett:
This PR is preparatory work for: https://www.pivotaltracker.com/story/show/93551518

This PR makes it possible to use Tracker.js with only one of the analytics codes (classic or universal). Tracker.js currently expects both a universal code and a classic code but we are switching off classic a few apps at a time so Tracker.js should be not expect both codes or need them in order to work.

This PR:

- normalises the interfaces of `google-analytics-classic-tracker.js` and `google-analytics-universal-tracker.js` so that `Tracker.js` can treat them both generically as trackers.
- Changes Tracker.js to accept one or both or none of the analytics ids.

@fofr it would be cool if you could take a look at this, but anyone else's reckons also welcome.

### DIFF
--- a/javascripts/govuk/analytics/tracker.js
+++ b/javascripts/govuk/analytics/tracker.js
@@ -16,7 +16,7 @@
   };
 
   Tracker.prototype.trackPageview = function(path, title) {
-    this.classic.trackPageview(path);
+    this.classic.trackPageview(path, title);
     this.universal.trackPageview(path, title);
   };
 
@@ -42,8 +42,8 @@
     Check this for your app before using this
    */
   Tracker.prototype.setDimension = function(index, value, name, scope) {
-    this.universal.setDimension(index, value);
-    this.classic.setCustomVariable(index, value, name, scope);
+    this.universal.setDimension(index, value, name, scope);
+    this.classic.setDimension(index, value, name, scope);
   };
 
   /*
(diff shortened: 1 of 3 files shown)


javascripts/govuk/analytics/tracker.js
```diff
@@ -6,18 +6,25 @@
...
+    }
+    if (typeof config.classicId != 'undefined') {
+      this.trackers.push(new GOVUK.GoogleAnalyticsClassicTracker(config.classicId, config.cookieDomain));
+    }
   };
 
   Tracker.load = function() {
-    GOVUK.GoogleAnalyticsClassicTracker.load();
-    GOVUK.GoogleAnalyticsUniversalTracker.load();
+    for (var i=0; i < this.trackers.length; i++) {
```
fofr: This method is called statically before a Tracker object is instantiated. So `this.trackers` won't exist at this stage.

alicebartlett: Ah yeah, I got a bit carried away here didn't i... will amend. Thanks!

fofr: Although this now means both the old and new GA libraries will always be loaded, even if they aren't needed.

fofr:
A summary of a conversation with @alicebartlett
The dual loading of both GA libraries, even if only classic or universal is being used, is a concern. Avoiding the extra http hit to GA is preferable, while the behaviour of `ga.js` when loaded without any configuration is unknown – eg it could set cookies, though it probably doesn't.

In the context of this story https://www.pivotaltracker.com/n/projects/1261204/stories/93551518 (testing a deprecation of classic analytics) – it seems acceptable to continue to load both.

PR #192, opened 2015-05-12
Add link to example table using tabular numbers

edwardhorsford:
This PR links to elements which includes an example table using tabular numbers.

I don't know if we link to elements in other cases, but it includes a lot of good use cases.


alicebartlett:
Thanks @edwardhorsford, sorry it took so long to merge this!

PR #194, opened 2015-05-21
Remove Google Analytics classic

fofr:
On 2 June we plan to deprecate Google Analytics classic on GOV.UK, from then on we will only use Universal analytics. Google Analytics classic data collection techniques are being deprecated by Google. We should no longer support classic and new tracking should use only Universal.

* Remove classic tracker and tests
* Remove classic specific method arguments (eg custom variable slot and name)
* No longer load the old `ga.js` library
* Keep an `Analytics` object as a wrapper for all analytics trackers.

GOVUK.Analytics was formerly a namespace used by slimmer. This has been removed and all apps using slimmer updated. Usage goes from:
`GOVUK.analytics = new GOVUK.Tracker(…)` to `GOVUK.analytics = new GOVUK.Analytics(…)`

Because the Tracker and TrackerSpec files have been renamed, when reviewing I suggest looking at the individual commits.

An example of how apps will update after this change:
https://github.com/alphagov/government-service-design-manual/compare/remove-classic-analytics

cc @edds @tombye @alicebartlett @benilovj


edds:
What is the rational behind moving from `Tracker` to `Analytics`? I thought the benefit of doing the original work was when we wanted to remove classic we could just remove the calls from the `Tracker` object and consuming applications wouldn't have to change.

fofr:
`Tracker` has been renamed because the object is a collective representation of all analytics. When originally written the slimmer GOVUK.Analytics namespace prevented this, although this was always the intention.

In any case, most consuming applications will need to remove their references to `google-analytics-classic-tracker.js`. It also makes sense to remove classic specific parameters at this point, eg name and scope values passed to `setDimension`.

fofr:
Is this good to merge?

edds:
Sorry, will have a look at the actual changes in a moment.

fofr:
Thanks @edds :smile:

PR #196, opened 2015-05-27
Proposal to fix JSLint Error for side effects

worktick:
I've been receiving a JSLint message "Do not use 'new' for side-effects." when used in this context:

```javascript
    new GOVUK.SelectionButtons($blockLabels);
```

This code is also available at: [MoJ - Accelerated Claims - line: 88](https://github.com/ministryofjustice/accelerated_claims/blob/34a625f713a0e488d8673db9bbc48d476c0e84a5/app/assets/javascripts/application.js.erb)

With the change I've made we would be able to achieve better LINTiness

```javascript
    var selectionButtons = new GOVUK.SelectionButtons();
    selectionButtons.$elms = $blockLabels;
    selectionButtons.setInitialState($blockLabels);
```

I would prefer a better pattern, such as an Initialise method (ideally chainable off the constructor) - but I'll take what I can get ;)

### DIFF
--- a/javascripts/govuk/selection-buttons.js
+++ b/javascripts/govuk/selection-buttons.js
@@ -19,7 +19,7 @@
       $elms = $(elmsOrSelector);
       this.selector = elmsOrSelector;
       this.setInitialState($(this.selector));
-    } else {
+    } else if ($elms !== undefined) {
       this.$elms = elmsOrSelector;
       this.setInitialState(this.$elms);
     }
--- a/javascripts/govuk/selection-buttons.js
+++ b/javascripts/govuk/selection-buttons.js
@@ -19,7 +19,7 @@
       $elms = $(elmsOrSelector);
       this.selector = elmsOrSelector;
       this.setInitialState($(this.selector));
-    } else if ($elms !== undefined) {
+    } else if (elmsOrSelector !== undefined) {
       this.$elms = elmsOrSelector;
       this.setInitialState(this.$elms);
     }


edds:
Thanks for proposing this change, I would be happy to get this merged so you can call the constructor without any arguments. Would you first be able to merge these two commits into one for me though, and provide a description in the commit message as to what JSLint errors you are fixing it for. That way a future developer will be able to find out why that if statement is the way it is.

Also another way to get around the lint errors would be to instead call:

```javascript
var selectionButtons = new GOVUK.SelectionButtons($blockLabels);
```

Rather than splitting out the call to the constructor and calling methods on it.

lennym:
I'm not sure I agree with this change, it doesn't really make sense to call this constructor without any context and it makes the implementation a lot more fragile.

I think changing it to require to call some presently encapsulated code outside of the constructor simply to satisfy a linter is a bad idea, since it increases the chance of creating breaking changes through changes to the internal implementation of the class. 

If the linting configuration you're not using is not working for you, then a better workaround would be to change the configuration, or to add a specific ignore flag for this case (or assign to a variable as @edds suggested).

worktick:
@edds
I agree with your suggestion to merge the changes and better define the Lint warning.
As for the suggestion to define a variable that (in my case and the MoJ example) would go unused, this would result in another Lint warning as well as being a bad practice.

@lennym
If the code was left "as is" then we have a situation where a constructor is adding event listeners and potentially modifying the DOM. In most other programming languages this is generally undesirable behaviour, hence the JSLint error "Do not use 'new' for side-effects".
Reporting this as a JSLint error side-steps any enforcement of my personal opinion and provides an objective reason that follows expected conventions.

lennym:
I've thought about this some more, and You are absolutely correct that "do not use new for side effects is a valid warning, but I think this is still the wrong way to approach a solution.

The warning is a result of the combined actions of instantiating an instance of a class whilst at the same time "doing things" (what jshint refers to as side effects). However in this case, the "doing things" (namely binding events listeners and DOM manipulation) is exactly what we intend to do, but instantiating a class is not. 

Hence the solution to the question of "do not use new for side effects" is not to leak some internals so that we are magically not having side effects, but instead to not use new, since we are more interested in the "side" effects than creating a reference to an instance of a class.

In my opinion, a better solution to this issue would be to refactor what is currently a constructor to instead be a (to couch it in design pattern speak) a factory function which returns an instance, but does not necessarily depend on the internal implementation being that of a constructor/prototype.

worktick:
I agree, given the intention of the code and the authors: the pattern of prototypal object construction is incorrect for this scenario. I was attempting to come at this from the angle of "let's not break backwards compatibility, but fix-up what's there".

By the way you refer to "$elms" and "setInitialState" as internals, in the pattern used there are better ways of preventing such API leakage.

Thanks for your feedback (and time) - I'll leave this for you guys to refactor

tombye:
I've copied the commits across from this with the changes suggested by @edds into https://github.com/alphagov/govuk_frontend_toolkit/pull/213 so closing to move the work across to that.

PR #197, opened 2015-06-02
Stop multiline from dropping below phase tag

alextea:
Setting `display: inline-block` was causing the span to drop below the phase tag. Also removed `width` declarations as inline elements can’t have width set explicitly.

I also think the `vertical-align` statements could be removed as they don't seem to have any visual effect (in Chrome).

## Before

screen shot 2015-06-02 at 15 29 34

## After

screen shot 2015-06-02 at 15 29 46

### DIFF
--- a/stylesheets/design-patterns/_alpha-beta.scss
+++ b/stylesheets/design-patterns/_alpha-beta.scss
@@ -27,12 +27,9 @@
   }
 
   span {
-    @include inline-block;
-    width: 75%;
     vertical-align: top;
 
     @include media(tablet) {
-      width: auto;
       vertical-align: baseline;
     }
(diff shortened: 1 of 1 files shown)


tijmenb:
@edds okay with this?

tijmenb:
@edds should we bump the gem version and update the changelog in this pull request, or do we do that in a separate PR?

edds:
@tijmenb we like separate pull requests for change log and version bumps.

dsingleton:
@tijmenb theres some doc on the version bump process, but it's not obvious: https://github.com/alphagov/styleguides/blob/master/rubygems.md#versioning

PR #200, opened 2015-06-11
Make heading mixins block elements

timpaul:
When you apply a heading mixin to a non-block element (like a form label) it should turn it into a block level element.

### DIFF
--- a/stylesheets/_typography.scss
+++ b/stylesheets/_typography.scss
@@ -130,6 +130,8 @@ $is-print: false !default;
 
 @mixin heading-80($tabular-numbers: false) {
   @include core-80($tabular-numbers: $tabular-numbers);
+  
+  display: block;
 
   padding-top: 8px;
   padding-bottom: 7px;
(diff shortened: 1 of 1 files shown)


alicebartlett:
[message from the frontend meeting] This seems like a reasonable change, could you provide a bit more background as to where this it's come from? 
We think it would be good to merge but as there is a possibility that this change will break the styles for some people's applications (as if they're using heading on a non-block level element it will then become block level), we'd like to hear more about it's origin.

timpaul:
Sure. There are some occasions when we'd like a form label to be visually represented as a heading of some kind (for example, very simple screens with only one form field). Form labels are inline by default, which is fine, but it means you have to explicitly set them as block if you apply a heading style. This got me thinking about whether you'd ever want a heading to be inline, and it feels like 'blockness' is a core property of headings.

dsingleton:
We should probably resume the discussion next FE meeting, but this seems reasonable to me.

This _may_ introduce unexpected behaviour for someone updating, so we should consider if this would be a `major` change - or if that's overkill.

robinwhittleton:
I think I’d be happy for this to go in, but yes, it’d need to be a breaking change. I wonder if we should go to version 1.0.0 with this?

robinwhittleton:
Had a discussion with @dsingleton and decided that this isn’t a breaking change, but will do a minor bump for the next release.

PR #202, opened 2015-06-25
Fix typography line-heights for mobile and tablet

henryhadlow:
Denominator in the line-height fraction should always match $font-size-640
Numerator should always be a multiple of 5 (by convention)

### DIFF
--- a/stylesheets/_typography.scss
+++ b/stylesheets/_typography.scss
@@ -61,14 +61,14 @@ $is-print: false !default;
   @include _core-font-generator($font-size, $font-size-640, $font-size-print, $line-height, $line-height-640, $tabular-numbers, $font-weight);
 }
 
-@mixin core-27($line-height: (30 / 27), $line-height-640: (20 / 18), $tabular-numbers: false, $font-weight: 400) {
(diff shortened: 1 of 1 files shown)


alextea:
Why?

alicebartlett:
@henryhadlow we're going to close this, let me know if you think this should be reopened. If you want to reopen it, it would be great if you could provide screen shots.

henryhadlow:
Hey. Missed Alex's previous comment, sorry.

The denominator should always match the $font-size-640 because you want the fraction to resolve back to the line height you are specifying.
For example, currently `core-27 $line-height-640` = 20/18 = 1.111
The browser multiplies this by the current type size (`$font-size-640`) to get the line-height in px.
In this case, 20*1.111 = 22.22px

My [proposed change](https://github.com/alphagov/govuk_frontend_toolkit/pull/202/files#diff-303e01b4cd177a96931f7890fdf3422d) here works out like this:
`core-27 $line-height-640` = 25/20 = 1.25
Then 20*1.25 = 25px

We set leading and vertical spacing in 5px increments by convention. That's why I changed some of the numerators. For example, see lines 66–147 in https://github.com/alphagov/govuk_elements/blob/master/public/sass/elements/_elements-typography.scss

PR #208, opened 2015-08-03
Create simplest download and external link tracking plugins

fofr:
https://trello.com/c/ihxFfLWH/70-create-simplest-external-link-and-download-tracker-medium

This is a minimum viable tracker. It will work as expected in Chrome and Firefox – browsers that support `navigator.sendBeacon`, about 30% of all users. On older browsers we will still send analytics events, and we'll receive some data, but not all events will make it to GA and the dataset will be incomplete. We will however be able to estimate what proportion of events we're missing based on the complete Chrome and Firefox datasets – those estimates can be used to prioritise and influence a more complex solutions.

A fully-formed tracker would include, in older browsers, a means of preventing the page from loading until an event has been successfully tracked. This degrades the user experience and the complexities introduce a greater risk of links not working. The data collected from this tracker will provide much needed context and visibility.

* Add tracking to all external links (any link beginning with `http` and not linking to the current host)
* Add tracking to download links that match a given selector

Before merging, I'd love to get some feedback on this approach. Particularly:
* Is the javascript fault tolerant, if something fails will the link continue to load?
* Are the link CSS selectors accurate, could things be missed or tracked incorrectly?
* Should download links and external links be made more explicit? eg classes, rel attributes
* Is setting a single listener on the `body` tag the most performant way to create the event listeners?


bradleywright:
I don't know if using the file extension is a sensible way to decide if something is an attachment.

For GOV.UK we have two distinct attachment URLs:

- things at assets.digital.etc
- things at /government/uploads

Given that those are both specific to us rather than applicable to the toolkit, would it be worth making the marchers configurable and passing them in via static? That would make this solution more generally applicable but also more accurate to our use cases.

fofr:
@bradleywright That sounds like a sensible suggestion. I think the external link trackers will also need some similar configuration, eg a link to assets.digital.cabinet... would be a false positive.

Closing while I make changes.

tombye:
This is great (to read anyway).

One question I have is whether jQuery should be used in the tracking code as there's no dependancy, as far as I can see, for it in the code at the moment.

fofr:
jQuery is useful for:
* efficient event delegation
* complex CSS selectors that use Sizzle rather than `querySelectorAll` (and also older browser fallbacks)

fofr:
@bradleywright I've updated the download tracker so that it takes a selector argument and I've removed the file extension regex check.

fofr:
In static we'd do the following:
```js
GOVUK.analyticsPlugins.downloadLinkTracker({
  selector: 'a[href*="/government/uploads"], a[href*="assets.digital.cabinet-office.gov.uk"]'
});
```

bradleywright:
Function signature looks much more generally useful :+1: thanks for doing that.

I'll let someone else (@edds maybe?) review the code and answer your specific questions about performance and resilience.

fofr:
I've done some basic tests of resilience and don't believe JavaScript errors will cause a link not to load.

@edds Are you happy for this to be merged?

PR #213, opened 2015-08-07
Proposal to fix JSLint Error for side effects

tombye:
The JSLint error is:

["Do not use 'new' for side-effects."](https://jslinterrors.com/do-not-use-new-for-side-effects).

This change allows you to store the resulting
instance in a variable by calling the constructor
without any arguments.

This kind of use would mean the setting of the
`$elms` property and the calling of the
`setInitialState` method would still be required
for all behaviours to work correctly, for example:

```
var selectionButtons = new GOVUK.SelectionButtons();
selectionButtons.$elms = $blockLabels;
selectionButtons.setInitialState($blockLabels);
```

Credit to [worktick](https://github.com/worktick) for this work (see [#196](https://github.com/alphagov/govuk_frontend_toolkit/pull/196)).

### DIFF
--- a/javascripts/govuk/selection-buttons.js
+++ b/javascripts/govuk/selection-buttons.js
@@ -19,7 +19,7 @@
       $elms = $(elmsOrSelector);
       this.selector = elmsOrSelector;
       this.setInitialState($(this.selector));
-    } else {
+    } else if (elmsOrSelector !== undefined) {
       this.$elms = elmsOrSelector;
       this.setInitialState(this.$elms);
     }


lennym:
I think this has been proposed before, and I had some reckons about it then...

This is IMO approaching the problem from the wrong angle. The linter error message "Do not use new for side effects" is a highlighting of the mixture of concerns between the intention of `new Constructor()` as a creation of an instance, and the things that that class does. That is that one function call is serving two orthogonal purposes - it creates an instance, and it does other things (highlighted by the fact that you subsequently don't use the instance for anything).

Now, *in this case* we don't really have any interest in the creation of an instance (since we don't subsequently use it), so the appropriate response is to not use `new` rather than leaking the internals of the class - namely the "side-effects", which are the things that we *are* interested in.

tombye:
To be clear, the intention of this pull is to meet the requirements stated by @edds in his [comment on the original pull request](https://github.com/alphagov/govuk_frontend_toolkit/pull/196#issuecomment-105896551).

Being the same code, it doesn't invalidate any of the points made on the original pull request (thanks for the contribution btw), just to move those changes into a form that could be merged.

PR #214, opened 2015-08-10
Switch download tracking to use events

fofr:
Events show a user intent, but are less likely to be interpreted incorrectly – it shows that a user clicked a link on a page, rather than the file was successfully downloaded and viewed.

It should avoid inflating our pageview numbers, prevent an increase in our cardinality and leave open a more complete pageview tracking solution from logs (eg for direct downloads).


edds:
:+1:

PR #218, opened 2015-08-13
Allow JavaScript error tracking to be filtered

fofr:
Don’t track javascript errors generated by files that don’t match a specified filter. eg external files included by browser plugins.

Non `gov.uk` error events are currently tracked ~40,000 times per day (30% of error events tracked).

Usage:
```js
GOVUK.analyticsPlugins.error({filenameMustMatch: /gov\.uk/});
```


edds:
Would it be better to do this in GA rather than doing it in code? That way we can still have a view with all the events incase it is useful in the future.

fofr:
Good suggestion @edds.

Speaking with Paul Cronk about the filters we'd need for this, and the volume of events we're collecting we are going to try doing this in GA first.

Closing PR

fofr:
Re-opening this PR. After investigating doing this in GA we've found it's not currently possible. You cannot filter both on an event action and an event label, and you cannot limit this to just one aspect of a profile. An emailed report could be generated, but this would be suitable for all use cases – eg a deploy where real time events are checked to see if JavaScript errors are affected.

Going back to the original approach.

PR #222, opened 2015-10-27
Add helpers partial for functions

gemmaleigh:
This PR creates a new partial `_helpers.scss`, and defines a pixel to em function.

It also adds documentation and three usage examples.


KushalP:
This is great :+1: 

I especially like the addition of documentation for functions in the toolkit.

PR #223, opened 2015-10-30
Version 4.4.0

gemmaleigh:
- Add helpers partial for functions
- Add px to em function and documentation


dsingleton:
@gemmaleigh @tijmenb it looks like the CI job to push new versions has [been failing](https://rubygems.org/gems/govuk_frontend_toolkit), so this wasn't released to [rubygems](https://rubygems.org/gems/govuk_frontend_toolkit) / etc

fofr:
I've run the task again, it finished successfully and the new versions have been published.

dsingleton:
It looks like the build is only failing on one of the CI hosts (`ci-slave-4`), and the rebuild passed as it ran on a different host - so this is likely to keep happening.

I've chatted to infrastructure about the CI inconsistency, which they'll investigate, but recommend setting an explicit `.ruby_version` for the app, which we should be doing anywhere. I'll open a PR for that.

dsingleton:
PR to set ruby version: https://github.com/alphagov/govuk_frontend_toolkit/pull/224

PR #226, opened 2015-11-11
Run tests against jQuery 1.11.3

fofr:
* Switch to the official jQuery package
* Use 1.11.3 which is the version currently used by GOV.UK
* Fix SelectionButton tests for jQuery 1.11.3
  * Checkbox and radio states should be set using `prop` rather than `attr`
  * Trigger reliably changes the state of a checkbox, we don’t need to set the `attr` as well – doing both is confusing and might achieve the opposite of what’s intended

This raises the issue of how we should communicate that:
* jQuery is required for using some of the JS
* Which version of jQuery we should nominally support, what's the minimum version
* Potentially running tests against a set of jQuery libraries

cc @robinwhittleton @dsingleton

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -29,7 +29,7 @@ module.exports = function(grunt) {
     jasmine: {
       javascripts: {
         src: [
-          'node_modules/jquery-browser/lib/jquery.js',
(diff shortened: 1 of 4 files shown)


robinwhittleton:
Good fix 👍

PR #227, opened 2015-11-12
Find and auto-start JavaScript modules from markup

fofr:
Find and auto-start modules specified using the `data-module=""` pattern in markup. Includes an example module for tracking Google Analytics events using markup, which will meet the need in this PR: https://github.com/alphagov/frontend/pull/894

This is based on the same pattern used by `govuk_admin_template` and all our publishing tools:
alphagov/govuk_admin_template#3 and https://github.com/alphagov/transition/pull/192

I've created an example of how this could be used on static:
https://github.com/alphagov/static/compare/example-module-usage

This addresses the problems of:
* needing to include `script` tags within views for specific JavaScript to run
* determining where and when to instantiate JavaScript ie in a `base.js` file, or in a view specific one
* JavaScript acting on an entire page and having side effects eg acting on all `.js-x` classes on a page, rather than within a specific part of a page

This pattern should work well with components that are used on Static.

cc @robinwhittleton @quis @dsingleton @tombye @jamiecobbett 
Also cc'ing @tyom, a non-GDS user of frontend toolkit.

## From the documentation

Javascript modules can be specified in markup using `data-` attributes:

```html
<div data-module="some-module">
  <strong>Some other markup inside the module</strong>
</div>
```

Modules can be found and started by including `govuk/modules.js` and running:

```javascript
$(document).ready(function(){
  GOVUK.modules.start();
});
```

This will attempt to find and start all modules in the page. For the example above it will look for a module at `GOVUK.Modules.SomeModule`. Note the value of the data attribute has been converted to _PascalCase_.

The module will be instantiated and then its `start` method called. The HTML element with the `data-module` attribute is passed as the first argument to the module. This limits modules to acting only within their containing elements.

```javascript
module = new GOVUK.Modules[type]();
module.start(element);
```

The simplest of modules looks like this:

```javascript
(function(Modules) {
  "use strict";
  Modules.SomeModule = function() {
    this.start = function(element) {
      // module code
    }
  };
})(window.GOVUK.Modules);
```

### Writing modules

Whilst this isn’t prescriptive, it helps if modules look and behave in a similar manner.

#### Use `js-` prefixed classes for interaction hooks

Make it clear where a javascript module will be applying behaviour:

```html
<div data-module="toggle-thing">
  <a href="/" class="js-toggle">Toggle</a>
  <div class="js-toggle-target">Target</div>
</div>
```

#### Declare event listeners at the start

Beginning with a set of event listeners clearly indicates the module’s intentions.

```js
this.start = function(element) {
  element.on('click', '.js-toggle', toggle);
  element.on('click', '.js-cancel', cancel);
}
```

Where possible, assign listeners to the module element to minimise the number of listeners and to allow for flexible markup:

```html
<div data-module="toggle-thing">
  <a href="/" class="js-toggle">This toggles</a>
  <div class="js-toggle-target">
    <p>Some content</p>
    <a href="/" class="js-toggle">This also toggles</a>
  </div>
</div>
```

#### Use data-attributes for configuration

Keep modules flexible by moving configuration to data attributes on the module’s element:

```html
<div
  data-module="html-stream"
  data-url="/some/endpoint"
  data-refresh-ms="5000">
  <!-- updates with content from end point -->
</div>
```

#### Include Jasmine specs

Modules should have their own tests, whether they’re being included with the gem or are app specific.


javascripts/govuk/modules.js
```diff
@@ -0,0 +1,56 @@
...
+      for (var i = 0, l = modules.length; i < l; i++) {
+        var module,
+            element = $(modules[i]),
+            type = camelCaseAndCapitalise(element.data('module')),
+            started = element.data('module-started');
+
+        if (typeof GOVUK.Modules[type] === "function" && !started) {
+          module = new GOVUK.Modules[type]();
+          module.start(element);
+          element.data('module-started', true);
```
quis: I’d like to see this documented as a reserved attribute.

quis:
Code thoughts: this is a good, solid foundation, and it’s not overengineered.

Higher level thoughts: standardising this wrapper will make it easier to contribute some of our team’s work to back to the toolkit.

Overall: :+1:

dsingleton:
:+1: to the approach / pattern. Adds some simple structure / scoping to our JS. It's also a stepping stone to doing something more clever in the future with lazy loading / `requirejs` strategies.

We discussed naming a bit offline, in particular `GOVUK.modules` vs `GOVUK.Modules` and `GOVUK.modules.start(container)`, but naming hard, and this certainly good enough to ship. If we decide to change it in the future we can with a major version bump.

Once this is merged it'd be great to use this in a few places, to make use of it, and so anyone wanting to use it has a few more points of reference. +:100: to using this with components, it's a great fit.

spec/unit/ModulesSpec.js
```diff
@@ -0,0 +1,89 @@
...
+
+      GOVUK.modules.start(container);
+
+      var args = callback.calls.argsFor(0);
+      expect(args[0].is('div[data-module="test-alert-module"]')).toBe(true);
+    });
+
+    it('starts all modules that are on the page', function() {
+      var modules = $(
+            '<div data-module="test-alert-module"></div>\
```
robinwhittleton: What’s our policy on dependency versions? If we’re happy with specifying a minimum Node version of 4.0 then we get ES6 template strings which are a lot nicer than the non-standard ‘\ = newline’ convention we’re following here. Minor nit though.

robinwhittleton:
Apart from that, only my own 👍 to add. Happy to merge if everyone else is.

javascripts/govuk/modules.js
fofr: Good point, I'll add this to the documentation.

spec/unit/ModulesSpec.js
fofr: I like to keep these specs backwards compatible with older browsers, so if necessary Jasmine tests can be run on older browsers to try and find problems.

robinwhittleton: Good call.

javascripts/govuk/modules.js
fofr: Updated docs in 2ab43b2

PR #228, opened 2015-11-13
Version 4.5.0

fofr:
- Find and auto-start JavaScript modules from markup:
`data-module="module-name"`(PR #227)


dsingleton:
Also includes the explicit setting of a ruby version
https://github.com/alphagov/govuk_frontend_toolkit/compare/v4.4.0...master, might be worth mentioning? Otherwise happy to merge.

dsingleton:
Talked to @fofr about the ruby bump and agreed it doesn't need to be in the changelog. 

The ruby version only affects the dev/testing part of this repo, not the things exported in packages, eg, Gem/NPM.

PR #230, opened 2015-11-19
Allow the toolkit font to be changed

fofr:
Some projects, such as petitions and admin tools aren’t GOVUK branded but still use the toolkit. These projects sometimes can’t use New Transport, or choose not to.

* Make it easy to switch out the font used
* Add documentation
* Write font stacks on single line as that’s how they are usually
presented in CSS, etc.

cc @quis @robinwhittleton

### DIFF
--- a/stylesheets/_font_stack.scss
+++ b/stylesheets/_font_stack.scss
@@ -1,6 +1,5 @@
 //  GOV.UK font stacks, referred to in typography.scss
 
-
 //  Font stack weirdness
 //
 //  To ensure embedded fonts fall back to appropriate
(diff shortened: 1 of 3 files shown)


dsingleton:
:100: :+1: to the idea. This will save quite a few projects having to outright fork the toolkit to make this change.

stylesheets/_font_stack.scss
```diff
@@ -35,10 +27,7 @@ $NTA-Light-Tabular:
        local("Helvetica");
 }
 
-$Helvetica-Regular:
-  "GDS-Logo",
-  sans-serif;
+$Helvetica-Regular: "GDS-Logo", sans-serif;
```
gemmaleigh: It might help to move this above, to sit with the other Sass variables for font stacks.

fofr: I think it needs to stay below the `@font-face` declaration, but I guess both blocks could move up? (There's no reference to this within the toolkit, but it's used in a couple of apps for organisation logo styles).

gemmaleigh: Should all the @font-face rules sit together? The rest are in the govuk template:
https://github.com/alphagov/govuk_template/blob/master/source/assets/stylesheets/fonts.css

alextea: @gemmaleigh I think the fonts aren't in here deliberately, to avoid licensing issues. I've just had a discussion about this on slack: https://ukgovernmentdigital.slack.com/archives/frontend/p1448023249000062

fofr:
Closing until I have addressed @robinwhittleton's feedback.

fofr:
I've rebased to remove conflicts with master and updated documentation to include @robinwhittleton’s changes.

dsingleton:
👍 

Longer term i've been thinking we make make all our variables overridable by default - possibly pulling them out in to a separate module for visibility (or using SASS doc to expose them)

PR #231, opened 2015-11-20
Rename tests from `SomeFileSpec.js` to `some-file.spec.js` to match source

fofr:
The source files are in the format: `some-file.js`, change the tests from `SomeFileSpec.js` to a similar `some-file.spec.js`.

When searching for a JS file both the source and the test can be easily found.


dsingleton:
I'm kind of indifferent to this. The change makes sense for development reasons, but camelCase feels a bit more JS-y (totally unqualified hot take there).

I'd be happy merging this. Plus it's a good example of coding/naming conventions from FET2 discussions (CC @robinwhittleton)

robinwhittleton:
I’m indifferent. On the basis of consistency at least 👍

PR #235, opened 2016-01-07
Add images from govuk elements

gemmaleigh:
This PR adds the remaining image from govuk elements which is used by the govuk elements Sass files.

1. ~~Add separator images for the breadcrumb divider~~ (these were added in #236).
2. Add a left pointing arrow image, as a fallback for the back link


gemmaleigh:
This PR adds the left pointing arrow image, [which is only used as a fallback](https://github.com/alphagov/govuk_elements/blob/master/public/sass/elements/_elements-typography.scss#L222) for IE8 and below.

fofr:
:+1:

PR #236, opened 2016-01-07
Add breadcrumb styles to design patterns

gemmaleigh:
These are duplicated in both alphagov/static and
alphagov/govuk_elements and aren’t yet used by govuk components.

Move these styles to the front end toolkit, to avoid variants of these
breadcrumb styles appearing in more than one place.

The separator images are been added to the images folder in #235.


stylesheets/design-patterns/_breadcrumb.scss
```diff
@@ -0,0 +1,43 @@
...
+    background-repeat: no-repeat;
+
+    list-style: none;
+
+    margin-right: 0.5em;
+    margin-bottom: 0.4em;
+    margin-left: 0;
+    padding-right: 1em;
+
+    &:last-child {
```
fofr: I'm working on a breadcrumb component in static, while working on it I spotted that this doesn't look great in IE8 and below because they dont support `last-child`, they do however support `first-child`.

See https://github.com/alphagov/static/compare/breadcrumb-component and https://github.com/alphagov/static/commit/dc60ba5f805bc0d9d8d0721cfe1e933ed2184d79

fofr:
As part of the component pattern, a file `breadcrumbs.scss` gets added to `static` and all styles constrained to the selector `.govuk-breadcrumbs { … }`.

I suppose that file could import the toolkit styles within the block.

gemmaleigh:
This is a now a mixin, so it won't output anything unless imported.

The govuk breadcrumb component could then import this mixin:

    .govuk-breadcrumbs {
        @import breadcrumbs;
    }


The [alpha label](https://github.com/alphagov/static/blob/master/app/assets/stylesheets/govuk-component/_alpha-label.scss#L2) and beta label components both import the phase banner mixin from the front end toolkit, in the same way.

cc. @fofr, @dsingleton.

fofr:
Awesome, that's what I was hoping for :smile:

Are the `separator.png` images already in the toolkit?

gemmaleigh:
Hi @fofr, I've added the separator images and update the docs. Is this good to merge?

cc. @dsingleton, @tombye.

PR #244, opened 2016-02-02
Add tracking of click events on mailto links

erikse:
Not sure on the best way to share functions between external-link-tracker.js and mailto-link-tracker.js, other than the category value passed to trackEvent, the two functions in each file are essentially identical


fofr:
This looks good to me :+1:

erikse:
Cool, maybe we can merge and give it a go? I don't have permissions

PR #250, opened 2016-02-16
Add a new Sass variable $footer-background

gemmaleigh:
Add a new Sass variable for the GOV.UK template's $footer-background.

The govuk_template uses `$grey-8` in `/styleguide/colours.scss`.

Since `$grey-8` is the same as `$grey-3` from the front end toolkit = `#dee0e2`, but `$grey-3` is overridden by the govuk template - in `/styleguide/colours.scss` (line 10), create a new variable $footer-background to be used by the template.


fofr:
This looks good to me :+1:

I'd like to hear what others think though.

PR #251, opened 2016-02-16
Pass cohort name to universal analytics set dimension

h-lame:
During testing of alphagov/smart-answers#2265 we noticed that only the test name was being sent to GA for our custom dimension where we expected to see the cohort name too.  After a bit of digging we found that the `MultivariateTest` had not been updated since we removed the classic GA tracker analytics implementation.  Old dimensions allowed 4 arguments: the dimension index, a name (this was the name of the test), a value (this was the name of the cohort), and a scope for the data to be set in (this was session scope); whereas new dimensions allow only 2 arguments: the dimension index and a piece of custom data (because all other arguments were ignored this was just the name of the test).

We want to pass on the name of the cohort to the dimension so we change the call to set dimension from the `MultivariateTest` to combine the test name and the cohort name and send this as the 2nd argument to the call.

While we're at it we update the contact details for the person to communicate with about creating new custom dimensions as it's no longer Ashraf.

/cc @benilovj, @fofr

### DIFF
--- a/docs/javascript.md
+++ b/docs/javascript.md
@@ -248,7 +248,7 @@ var test = new GOVUK.MultivariateTest({
 });
 ```
 
-`customDimensionIndex` is the index of the custom variable in Google Analytics. GA only gives 50 integer slots to each account, and it is important that a unique integer is assigned to each test. Current contact for assigning a custom var slot for GOV.UK is: Ashraf Chohan <ashraf.chohan@digital.cabinet-office.gov.uk>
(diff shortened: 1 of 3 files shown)


h-lame:
FWIW: I have tested this locally with the GA chrome debugger and it does send what I expect now.  Showed it to Tim LB and he seemed happy with that as a test.

fofr:
Looks good to me :+1:

PR #252, opened 2016-02-17
Bump version to 4.8.0

h-lame:
I'd like to get #251 out so I can redo the smart-answers PR (alphagov/smart-answers#2265) that revealed the problem.


VERSION.txt
```diff
@@ -1 +1 @@
-4.7.0
+4.8.0
```
fofr: Shouldn't this be 4.7.1 as it's a bug fix?

h-lame: While it is a bug fix, it also changes how we communicate with GA.  In the off chance anyone was using and relying on it, they'd need to redo their analytics, so it felt "bigger" to me than a simple bug fix.

fofr:
:+1:

PR #254, opened 2016-02-18
Version 4.8.1

fofr:
Update DEFRA brand colour to new green (PR #249)


benlovell:
Is this classed as a bug-fix?

jamiecobbett:
@benlovell I thought about that and decided that a colour change doesn't really fit into semver. Thinking about it again, I suppose you could call it a MINOR because we're "adding functionality in a backwards-compatible manner".

benlovell:
@jamiecobbett It's no biggy. :+1:

PR #258, opened 2016-02-25
Remove alias for footer background

gemmaleigh:
This Sass variable `$footer-background` should remain with the [GOV.UK template’s colour
variables](https://github.com/alphagov/govuk_template/pull/201/files).

cc. @dsingleton.


dsingleton:
:+1: 

We agreed we're going to start splitting non-template specific helpers out into the toolkit, but for anything specific to template, eg this grey is specifically for use with templates' footer, they'll stay in template.

PR #259, opened 2016-02-26
Update scss-lint to match govuk-lint

gemmaleigh:
- Exclude the CSS3 partial, as before
- Set max_depth for nesting to 3, which is govuk-lint’s ideal, it is
currently set to 5, but we can achieve 3 here.

cc. @dsingleton, @fofr.

### DIFF
--- a/.scss-lint.yaml
+++ b/.scss-lint.yaml
@@ -1,32 +1,193 @@
+# https://github.com/alphagov/govuk-lint/blob/master/configs/scss_lint/gds-sass-styleguide.yml
+
+# Default severity of all linters.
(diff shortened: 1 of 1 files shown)


dsingleton:
Just chatted offline and agreed it would be better to use the `govuk-lint` gem directly instead.

It'll avoid the linting configuration being duplicated, and possibly going out of sync. There's a good example of adding `govuk-lint` to a project here: https://github.com/alphagov/government-frontend/pull/101

gemmaleigh:
Closing in favour of using the `govuk-lint` gem in #260.

PR #260, opened 2016-02-26
Use GOV.UK lint for scss linting

gemmaleigh:
Use the [GOV.UK lint](https://github.com/alphagov/govuk-lint) gem.

Remove grunt-scss-lint, instead use grunt-shell

Use grunt-shell to run `bundle`, then `bundle exec govuk-lint-sass stylesheets` to lint the scss files in `/stylesheets`.

Fix scss lint offences

Changes made:

Change the focus colour for the media player to use the $focus-colour variable, rather than the CSS color orange. 

Swatch below to compare (orange on the left, `$focus-colour` = `$yellow` on the right). 

orange-vs-focus copy

cc. @dsingleton, @fofr.


Gemfile
```diff
@@ -1,4 +1,5 @@
 source 'https://rubygems.org'
 
-gem 'sass', '3.4.9'
-gem 'scss-lint', '0.30.0'
+gem 'sass', '3.4.15'
+gem 'scss_lint', '0.44.0'
```
fofr: Do we need to explicitly include `scss_lint`?

stylesheets/_typography.scss
```diff
@@ -15,6 +15,8 @@
 
 // CORE FONTS - NEW TRANSPORT
 
+// scss-lint:disable NameFormat
```
fofr: Is this being disabled so that `_something-` is acceptable? I think there's a provision in the linter to allow hyphenated names that start with an `_`. Though it looks like that's enabled: https://github.com/alphagov/govuk-lint/blob/master/configs/scss_lint/gds-sass-styleguide.yml#L94

stylesheets/design-patterns/_media-player.scss
```diff
@@ -222,7 +227,9 @@
         height: 100%;
         padding-bottom: 1px;
         width: 20px;
-        color: rgba(0,0,0,0.5);
+        // scss-lint:disable ColorVariable
+        color: rgba(0, 0, 0, 0.5);
+        // scss-lint:enable ColorVariable
```
fofr: I'm not sure this needs to be re-enabled, when disabled within a block I _believe_ the linter is disabled for just that block: https://github.com/brigade/scss-lint (see "Disable all lints within a block (and all contained blocks)")

stylesheets/_font_stack.scss
```diff
@@ -10,10 +10,12 @@
 //  the font-stack in a font-face declaration rather
 //  than with the usual font-family.
 
+// scss-lint:disable NameFormat
```
fofr: It'd be good to include a reason with each disabled linter. I'd like to enable this linter in the future: https://github.com/brigade/scss-lint/blob/master/lib/scss_lint/linter/README.md#disablelinterreason

fofr:
:heart: This was on my todo list, really happy to see this, and some good fixes too.

stylesheets/_typography.scss
gemmaleigh: The requirement is that all variables names are lower case and hypenated. For the font stack variables, these have uppercase letters, eg `$NTA-Light` and `$Helvetica-Regular`.

I've disabled the linter to allow these variable names in the font-stack and typography partials.

stylesheets/design-patterns/_media-player.scss
gemmaleigh: You're right, it shouldn't need the closing comment - I've removed it.

stylesheets/_font_stack.scss
gemmaleigh: I've added reasons for each, having a context for each disabled linter is really useful.

stylesheets/design-patterns/_media-player.scss
```diff
@@ -222,7 +227,9 @@
         height: 100%;
         padding-bottom: 1px;
         width: 20px;
-        color: rgba(0,0,0,0.5);
+        // Allow RGBA here, it is the only colour value with alpha transparency
+        // scss-lint:disable ColorVariable
+        color: rgba(0, 0, 0, 0.5);
```
dsingleton: I had a similar linting issue on [static](https://github.com/alphagov/static/pull/745/files#diff-17beb2dcf03801a29d1ac8f73879ca20R434) and used SCSS's `transparentize` with `$black` to meet linting without needing to disable a rule.

dsingleton:
This is great :clap:

PR #262, opened 2016-03-02
Add `websafe` departmental colours, ported from Whitehall

fofr:
We use `websafe` to mean strong enough contrast against white to be used for copy and meet the AAA (large text) and AA (smaller copy) WCAG guidelines.

* Port websafe departmental colours from Whitehall (original: https://github.com/alphagov/whitehall/blob/540b900378823c2aa7b3baf3c524528dd48ee30e/app/assets/stylesheets/frontend/styleguide/_colours.scss)
* Include `$all-organisation-brand-colours` list with example usage
* Split palette and organisation colours into two files to maintain backwards compatibility (Import order matters as organisations uses `link-color` which is defined in palette.)

I'd like to get feedback on whether this file structure makes sense. One problem might be, a consuming app that imports _just_ `organisations` will get a `$link-color not defined` error.

cc @gemmaleigh @dsingleton @markhurrell


boffbowsh:
Bump and /cc @alextea for an additional reviewer. It LGTM but Paul wants opinions on the new structure

alextea:
Structure makes sense to me. :+1: 

Is this a breaking change? Or would the problem @fofr described only affect new apps?

boffbowsh:
Older apps currently import `colours` which now does the imports in the correct order, so that's fine. New apps would need to ensure they have the imports in the correct order if they import `organisations`

alextea:
Great, I don't think that's a blocker.

fofr:
:+1:

PR #264, opened 2016-03-03
Improve phase banner text wrapping

alextea:
Previously the text block wrapped below the phase tag, and the ascenders were bumping up against the phase tag.

The change means the text wraps under itself, avoiding the phase tag.

Before:
screen shot 2016-03-03 at 10 55 34

After:
screen shot 2016-03-03 at 10 54 40

### DIFF
--- a/stylesheets/design-patterns/_alpha-beta.scss
+++ b/stylesheets/design-patterns/_alpha-beta.scss
@@ -22,6 +22,7 @@
     margin: 0;
     color: $banner-text-colour;
     @include core-16;
+    @extend %contain-floats;
   }
 
   .phase-tag {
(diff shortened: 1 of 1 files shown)


Davidslv:
:+1:

andymantell:
I like this in principle, however I feel that we do need to consider _slightly_ longer text. At a 320px wide viewport we are only 1 word away from a strange looking indent like this:

image

I know we talked about the hanging indent on slack requiring an extra container to apply the overflow: hidden to. I assume there's some resistance to changing the markup simply because it isn't easily rolled out to people without extra effort on their part?

andymantell:
This is what I had gone with on the Land Registry pattern library. By increasing the line height on the paragraph to 1.5 on mobile you get this:

image

robinwhittleton:
Duplicate of https://github.com/alphagov/govuk_frontend_toolkit/pull/266 - closing.

Issue #263, opened 2016-03-02
Line height too tight around the BETA / ALPHA tags in phase banners at mobile sizes


andymantell:
As reported in LandRegistry/land-registry-elements#11 the line height around the phase tags is a bit tight at mobile sizes, so when the text wraps to a new line it butts up against the bottom of the phase tag. See screenshot for example:

![image](https://cloud.githubusercontent.com/assets/134642/13468857/ad4f8eb0-e09c-11e5-840c-40b25a56a4d8.png)

alextea:
Hmmm, I vaguely remember fixing this a long time ago, so that the text wraps under "This". I will investigate!

PR #266, opened 2016-03-10
Improve phase banner alignment

gemmaleigh:
At smaller screen sizes, the phase banner text wraps underneath the phase tag.

Screenshot before:

gov uk elements - phase banner

This PR fixes this, by setting the parent paragraph to `display:table` and the line of text next to the phase tag to `display:table-cell`.

An alternative would be to use flexbox (as suggested by @alextea):

```
  p {
    display: flex;
    align-items: baseline;
  }
```
IE10 requires a prefix, uses a different syntax and doesn't work as well.

Without modifying the GOV.UK template, there isn't a way to target versions of IE greater than 8, or a way to test for a browsers' features.

Screenshots after:

iPhone 6

realios_iphone_6

Windows 10, IE11

win10_ie_11 0

Windows 7, IE8

win7_ie_8 0

### DIFF
--- a/stylesheets/design-patterns/_alpha-beta.scss
+++ b/stylesheets/design-patterns/_alpha-beta.scss
@@ -29,11 +29,7 @@
   }
 
   span {
-    vertical-align: top;
-
-    @include media(tablet) {
-      vertical-align: baseline;
-    }
+    vertical-align: baseline;
   }
 }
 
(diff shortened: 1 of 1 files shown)


alextea:
:clap:

gemmaleigh:
@alextea can you please close #264 if you're happy with this solution?

cc. @robinwhittleton.

robinwhittleton:
I suspect this won’t work in IE7, but then again it’s not going to look any worse that in does currently and it’s not on our testing schedule any more so 👍

andymantell:
Worth noting that this fixes #263

PR #269, opened 2016-04-15
Set default flag on font variables

daveraine:
I've had an [issue open](https://github.com/alphagov/govuk_template/issues/207) on govuk_template regarding the NTA font not containing international characters. I've been provided with a new font, "nta-extended", which contains a bigger character set. However, I can't override the font in the toolkit as when importing [_typography.scss](https://github.com/alphagov/govuk_frontend_toolkit/blob/master/stylesheets/_typography.scss) it imports font_stack, setting the font back to the standard "nta". By using the `!default` flag, the font can be overwritten before importing the toolkit sass files, and it won't be set back to "nta".

### DIFF
--- a/stylesheets/_font_stack.scss
+++ b/stylesheets/_font_stack.scss
@@ -18,7 +18,8 @@
 $NTA-Light:
   "nta",
   Arial,
-  sans-serif;
+  sans-serif
+  !default;
 
 // New Transport Light with Tabular
 
@@ -26,7 +27,8 @@ $NTA-Light-Tabular:
   "ntatabularnumbers",
   "nta",
   Arial,
-  sans-serif;
+  sans-serif
+  !default;
 
 // Helvetica Regular
 


fofr:
Adding for context:

This is similar to a PR I raised but didn't follow up on: https://github.com/alphagov/govuk_frontend_toolkit/pull/230

The difference with this branch is a change in name of the font-stack variable to be a bit more explicit, which becomes the thing that apps can edit.

fofr:
Is there a way we can include `nta-extended` in the toolkit, as an opt-in?

antimega:
@fofr it would be good to get nta-extended into the toolkit, but I'm out of date with how the font gets pulled in, and from where. I've got the font file, so suggestions and help welcome.

tombye:
I'd advocate the approach in [#230](https://github.com/alphagov/govuk_frontend_toolkit/pull/230).

The fonts are kept in the [govuk_template](https://github.com/alphagov/govuk_template/tree/master/source/assets/stylesheets) as `.eot` files and `.woff`s (which are embedded into the stylesheet with base64) so we'd need to add `nta-extended` there I think.

daveraine:
#230 looks like a better solution. Sorry, I didn't check through the closed PRs before opening this one.

fofr:
I've re-opened #230, closing this one in lieu.

Thank you for prompting this @daveraine :smile:

daveraine:
@fofr Awesome, thanks :+1:

PR #270, opened 2016-04-18
Version 4.10.0

fofr:
Allow New Transport font stack to be overridden by apps (PR #230)

> Some projects, such as petitions and admin tools aren’t GOVUK branded but still use the toolkit. These projects sometimes can’t use New Transport, or choose not to.

> * Make it easy to switch out the font used
* Add documentation
* Write font stacks on single line as that’s how they are usually
presented in CSS, etc.

Also allows apps to use `nta-extended`, see #269 and https://github.com/alphagov/govuk_template/issues/207 for reference.


gemmaleigh:
👍

fofr:
@daveraine The gem is now versioned so you can give it a try. Please let us know if this works for you.

daveraine:
Working perfectly, thanks.

PR #271, opened 2016-04-26
Add support for font-size-adjust into the font stack

robinwhittleton:
`font-size-adjust` allows you to amend the aspect value of fallback fonts to match your original font: https://www.w3.org/TR/css-fonts-3/#font-size-adjust-prop. A specific use case for us would be getting foreign characters that are not in NTA to show up with matching metrics.

For NTA the original aspect ratios are 0.5 for regular text and 0.525 for bold (this can be tested by adding a `font-size-adjust` to gov.uk and adjusting the property until no difference is seen when it’s toggled). Applying these values makes the fallback of Arial match NTA’s metrics.

Current support is Firefox (and Chrome behind a flag). Testing shows minor changes, but it could be useful in certain situations, for example:

NTA (with or without `font-size-adjust`):
nta

Arial with `font-size-adjust`:
arial_with_fsa

Arial without `font-size-adjust`:
arial_without_fsa

### DIFF
--- a/stylesheets/_typography.scss
+++ b/stylesheets/_typography.scss
@@ -33,6 +33,14 @@ $is-print: false !default;
   font-weight: $font-weight;
   text-transform: none;
 
+  @if $toolkit-font-stack == $NTA-Light {
+    @if $font-weight > 400 {
+      font-size-adjust: 0.525;
+    } @else {
+      font-size-adjust: 0.5;
+    }
+  }
+
   @include media(tablet) {
     font-size: $font-size;
     line-height: $line-height;


fofr:
Looks good to me :+1:. Do you have equivalent screenshots for the bold variant?

robinwhittleton:
Oh yeah, knew I’d forgotten something. There’s actually no difference either way with the bold weight, but we need to include it in case the bold text is down the cascade from a `font-size-adjust: 0.5`.

NTA with `font-size-adjust`:
nta

Arial with `font-size-adjust`:
arial_with_fsa

Arial without `font-size-adjust`:
arial_without_fsa

fofr:
:+1:

Issue #229, opened 2015-11-16
Viewport rules repeated in generated CSS


fofr:
The rules found at https://github.com/alphagov/govuk_frontend_toolkit/blob/master/stylesheets/_conditionals.scss#L28-L34 are being repeated every time `_conditionals` is used.

eg Search for `-ms-viewport` in https://assets.digital.cabinet-office.gov.uk/static/govuk-template-b5d680ad6607cf5441a22a3a533e92b40b15b8f6b9708dbdfc5fc6da71949113.css

```css
@-ms-viewport {
  width: device-width;
}

@-o-viewport {
  width: device-width;
}
```

jonhurrell:
We've mitigated this issue on a project build by piping our sass stream through [node-sass-import-once](https://github.com/at-import/node-sass-import-once).

fofr:
Fixed by #272

PR #272, opened 2016-05-15
Stop CSS being output into a consuming project simply by including partials

robinwhittleton:
govuk_frontend_toolkit currently outputs a few declarations when you include the files. These can either be removed or migrated to govuk_template. Specifically, this PR removes the @font-face declaration for Helvetica, and removes the @viewport declarations (they’re added back in to govuk_template in https://github.com/alphagov/govuk_template/pull/212).

### DIFF
--- a/stylesheets/_conditionals.scss
+++ b/stylesheets/_conditionals.scss
@@ -25,14 +25,6 @@
 $is-ie: false !default;
 $mobile-ie6: true !default;
 
-@-ms-viewport {
-  width: device-width;
-}
-
-@-o-viewport {
-  width: device-width;
-}
-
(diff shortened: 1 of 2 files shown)


fofr:
This will also fix #229

fofr:
This and alphagov/govuk_template#212 both look good to me :+1:

Nice work.

fofr:
How do you plan on versioning this?

robinwhittleton:
Going by semver, it’s arguable. I think 4.11.0: removing the viewport statements would only break IE10 snap mode if https://github.com/alphagov/govuk_template isn’t upgraded at the same time, and the `GDS-Logo` @font-face was effectively a private API that should have been consumed by the `$HelveticaRegular` variable.

PR #273, opened 2016-05-17
WIP: Redefine $link-colour for better contrast

nickcolley:
Part of the A11Y Hack day, originally raised by Richard Morton.

Issue is present with the contrast between the focus style background-color and the default link color.

The change is subtle compared with $govuk-blue but passes WCAG 2.0 Level AA

Before
old

After
new

### DIFF
--- a/stylesheets/colours/_palette.scss
+++ b/stylesheets/colours/_palette.scss
@@ -55,7 +55,7 @@ $grey-4: #f8f8f8;
 $white: #fff;
 
 // Semantic colour names
-$link-colour: $govuk-blue;
+$link-colour: #00579a; // $govuk-blue with better contrast ratio
 $link-active-colour: $light-blue;
 $link-hover-colour: $light-blue;
 $link-visited-colour: #4c2c92;


dsingleton:
@nickcolley and I had a chat about this offline and decided there's a better approach. Rather than changing all link colours (slightly), we should add an explicit `link-focus-colour` variable and tweak that specifically.

Changing the global link colour has the downsides
- It means `link-colour` and `govuk-blue` diverge, which could potentially cause some confusion
- It tweaks the global link colour, which would need some more detailed contrast testing against different backgrounds, in addition to white/focus-oranage

Specifically targeting focus link colour lets us be explicit, and avoid those two problems :)

nickcolley:
One problem with having a specific focus colour is that the visited link colour is no longer reflected when a user focuses.

One solution would be to have two focus colours, intending on returning to this when I have more autonomy in the codebase.

selfthinker:
I was just looking for when our focussed link colour changed and only found this issue and it took ages to find the real change. Just in case someone else is looking for the same thing, this got fixed in https://github.com/alphagov/govuk_template/pull/272
(The reason why I couldn't find it was because I was looking for `#00579a` and the change uses `darken( $link-colour, 2.5%)`. I wonder if it's worth adding a comment to the file which references the hex colour?)

PR #275, opened 2016-05-20
Consistent button padding

aliuk2012:
This padding was originally set in GOVUK Elements button class instead it should have been be referenced in GOVUK Frontend Toolkits Button mixin.

Original issue: https://github.com/alphagov/govuk_elements/issues/228
GOVUK Elements PR: https://github.com/alphagov/govuk_elements/pull/230

### DIFF
--- a/stylesheets/design-patterns/_buttons.scss
+++ b/stylesheets/design-patterns/_buttons.scss
@@ -27,7 +27,7 @@
   // Size and shape
   position: relative;
   @include inline-block;
-  padding: 0.3em 0.6em 0.2em;
+  padding: .526315em .789473em .263157em; // 10px 15px 5px
   border: none;
   @include border-radius(0);
   -webkit-appearance: none;


gemmaleigh:
This won't work as the toolkit doesn't have access to the em helper function that GOV.UK elements does.

Since elements uses a base of 19px, these values are:

`padding: 0.5263157895em 0.7894736842em 0.2631578947em;`

screen shot 2016-05-20 at 15 19 00

Alternatively, there's a helpers file in the [frontend toolkit which has a pem function](https://github.com/alphagov/govuk_frontend_toolkit/blob/master/stylesheets/_helpers.scss), which would allow you to do this:

`padding: pem(10/19) pem(15/19) pem(5/19);`

Before making any changes, @dsingleton do you have an opinion on this?

I'd be keen to have one px-to-em helper function, which sits in the toolkit and I can then remove the one in GOV.UK elements.

dsingleton:
Overall 👍 to the change and moving changes upstream.

I had a chat with @gemmaleigh and decided it'd be best to make the change with regular unit values (eg, `0.526em`, rather than using the `em`/`pem` helpers right now. Using those in FET is something we'd want to look at holistically - and shouldn't hold up this change.

I'm surprised the tests passed with the reference to the missing `pem` helper :/

gemmaleigh:
@aliuk2012 we're going to put the decision about the px-to-em function into our backlog.

 Would it be possible to amend this PR to use these values for now?

`padding: .526315em .789473em .263157em; // 10px 15px 5px`

aliuk2012:
@gemmaleigh Sure I'll amend the PR to those values. 

@dsingleton speaking of tests I see that the Travis CI tests for both GE and FET do not appear to actually run the tests that mentioned in the README.

aliuk2012:
@gemmaleigh The changes are now in this PR.

gemmaleigh:
@aliuk2012 thanks for these changes. 

I'll bump the version once these have been merged, to then update GOV.UK elements.

dsingleton:
This looks good to me 👍  

This is being updated to match the definition in GOV.UK Elements, so Elements - and consumers of Elements won't see any difference. Any direct consumers of FET (like GOV.UK* itself, maybe Digital Marketplace?) will have a visible change. We'll need to make sure the `CHANGELOG` is clear for when those repos update.

Here's a preview of the change on GOV.UK
| Before | After | 
| --- | --- | 
|  before | after button |

> @dsingleton speaking of tests I see that the Travis CI tests for both GE and FET do not appear to actually run the tests that mentioned in the README.

They [do](https://travis-ci.org/alphagov/govuk_frontend_toolkit/builds/131731209) - but it's not very explicit. For a node project Travis will run `npm install` and `npm test` automatically. (same for `govuk_elements`)

gemmaleigh:
It looks like [this will affect Digital Marketplace](http://alphagov.github.io/digitalmarketplace-frontend-toolkit/forms/single-question-page.html) but those buttons are larger anyway as they are using 24px button text. 

@quis @tombye, are there other places where this is likely to be a problem?

PR #276, opened 2016-05-20
Ignore the automatically genererated SASS test files

dsingleton:
To test for syntax errors in the `scss` files we generate a temp
SASS file which includes all our scss files in `stylesheets/`, which
catches syntax errors in the top level declarations of those files.

Ignore the directory the tmp file is created in, not just the output.
Otherwise the test file/dir exists after running `npm test` and
could be accidentally commited.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,4 @@
 node_modules/
 .sass-cache
 npm-debug.log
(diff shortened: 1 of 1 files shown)


nickcolley:
LGTM 👍

PR #278, opened 2016-05-20
Bump version to 4.11.1

gemmaleigh:
# 4.11.1

- Increase button padding to match padding from GOV.UK elements (PR
#275)

This release increases the button padding, here are two screenshots to show how this has changed.

Before:

image

After:
image


gemmaleigh:
cc @robinwhittleton, @dsingleton.

dsingleton:
I think this should be a `MINOR` change, rather than a `PATCH. Applying SemVer to this project is tricky, as it's not an explicit API change, but it's also not backwards compatible - the button size change could have an effect on a consuming app.

A `MINOR` change gives a bit more warning to someone upgrading, and `MAJOR` would be overkill - we'll have lots of these kinds of changes as we consolidate things, and we'd end up approaching `99.x.y`.

dsingleton:
Makes me think we should try and document our application of SemVer to this repo (and similar changes on `govuk_template`, `govuk_elements`), to A) find consensus amongst ourselves B) have something written down for contributors and consumers. Thoughts?

gemmaleigh:
Closing in favour of bumping the version to 4.12.0 in #279.

PR #279, opened 2016-05-23
Bump version to 4.12.0

gemmaleigh:
Increase button padding to match padding from GOV.UK elements (PR #275).

If you have UI which depends on the padding set by the button mixin in
the frontend toolkit and this is not overridden by button padding set
by GOV.UK elements, this change will affect it.

This release increases the button padding, here are two screenshots to show how this has changed.

Before:

image

After:
image


gemmaleigh:
cc. @robinwhittleton.

PR #282, opened 2016-06-02
Update jquery object names to use a $ prefix

### LINKED ISSUE #241: update jquery object names to use a $ prefix
In the section on [JavaScript modules](https://github.com/alphagov/govuk_frontend_toolkit/blob/master/docs/javascript.md#usage), the variables `container` and `element` look like they are jQuery objects.

In the [style guide](https://github.com/alphagov/styleguides/blob/master/js.md#jquery) the guidance is to use $ prefixes to denote jQuery objects. Should we update the guidance here to reflect that?

gemmaleigh:
The variables container and element are jQuery objects, update the
guidance here to use the $ prefix as recommended by the style guide.

This fixes #241.

### DIFF
--- a/docs/javascript.md
+++ b/docs/javascript.md
@@ -80,8 +80,8 @@ Beginning with a set of event listeners clearly indicates the module’s intenti
 
 ```js
 this.start = function($element) {
-  element.on('click', '.js-toggle', toggle);
-  element.on('click', '.js-cancel', cancel);
+  $element.on('click', '.js-toggle', toggle);
+  $element.on('click', '.js-cancel', cancel);
 }
 ```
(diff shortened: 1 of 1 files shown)


docs/javascript.md
```diff
@@ -79,7 +79,7 @@ Make it clear where a javascript module will be applying behaviour:
 Beginning with a set of event listeners clearly indicates the module’s intentions.
 
 ```js
-this.start = function(element) {
+this.start = function($element) {
   element.on('click', '.js-toggle', toggle);
```
joelanman: these `element`s look like they need to be `$element` too?

PR #283, opened 2016-06-06
Update docs for IE conditionals

### LINKED ISSUE #280: Setting styles for IE8> using IE Helper Conditionals
So I've created an IE8 stylesheet which is linked via a conditional comment, am I correct in setting the flags, then importing in my main stylesheet again?
```
$is-ie: true;
$ie-version: 8;
@import 'style';
```

However, this is going to request an additional download of the duplicate rule-set of the main stylesheet, with the IE specific output. This obviously isn't great for performance. Is this the intended method?

gemmaleigh:
Add an example of conditional comments wrapping the main stylesheet
and conditional IE stylesheets.

Link to examples in the GOV.UK template and GOV.UK elements.

@jonhurrell, let me know if this would have resolved #280.
(I know it's closed, but think the updates will benefit others).

### DIFF
--- a/docs/mixins.md
+++ b/docs/mixins.md
@@ -67,16 +67,24 @@ edge of smaller screens.
 Media query and IE helpers. These make producing responsive layouts and
 attaching IE specific styles to elements really easy.
 
-To use the IE conditionals you will need to add extra stylesheets for each IE which look like:
+To use the IE conditionals you will need to add extra stylesheets for each IE,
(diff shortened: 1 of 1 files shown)


jonhurrell:
@gemmaleigh Yes, some examples would be great, it's a hard one to spot inline.

dsingleton:
In the future I was thinking we might want to split out some of the documentation into guides/tutorials, rather than being inline with individual mixins, eg, "How GOVUK Template supports internet explorer" or "Setting up GOVUK Template with Rails".

Thats quite a bit of work, and will probably come with larger refactoring of `govuk_template` and co, but this is a big improvement in the mean time 👍

PR #284, opened 2016-06-19
Add js for finding and starting govuk modules

gemmaleigh:
This is [being used by GOV.UK elements](https://github.com/alphagov/govuk_elements/pull/183/commits/22f9c130c5741daa06da02220895040d21c9d0e0), it _feels_ like it should sit in the front-end toolkit, as we have an example of this code here too. 

For now, this file adds javascript to [match the example js from the docs](https://github.com/alphagov/govuk_frontend_toolkit/blob/master/docs/java
script.md#modules).

### DIFF
--- a/javascripts/govuk/start-modules.js
+++ b/javascripts/govuk/start-modules.js
@@ -1,0 +1,5 @@
+// https://github.com/alphagov/govuk_frontend_toolkit/blob/master/docs/javascript.md#modules
+
+$(document).ready(function(){
+  GOVUK.modules.start();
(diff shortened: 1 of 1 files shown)


fofr:
This would be pulled in automatically by a lot of the ruby projects using the gem, and so it would run automatically. This goes against the rules of the toolkit, which lets you include it safely and then optin to what you need.

gemmaleigh:
I'll leave it where it is in that case, thanks @fofr  😄

PR #291, opened 2016-07-05
Add assignment of jQuery explicitly so it's consistent with other modules

nickcolley:
Follow up to https://github.com/alphagov/govuk_frontend_toolkit/pull/290#issuecomment-230437088

cc @robinwhittleton


colinrotherham:
For the record, perhaps note that this is in preparation for a future:

```javascript
var $ = require('jQuery');
```

i.e. Avoid globals.

PR #292, opened 2016-07-13
Configurable duration (in days) for AB Test cookie.

ikennaokpala:
Here, I have added a new option (cookieDuration), this option help to set the cookie expiration duration in days.

When this option isn't supplied the the default cookie duration is set to expire in 30 days.

## Motivation 

The motivation here is, by reducing the cookie expiration will allow returning users to be assigned
to different cohorts during A/B tests.

According to the performance analyst, A/B tests performance is measured using sessions and with
a default 30 day expiration, they will only see the one variant for that period.  Being able to 
reduce this increases the likelihood of users seeing different variants across their sessions.

### Sample code
```javascript
  new GOVUK.MultivariateTest({
      el: '.get-started a',
      name: 'startButton_201607',
      customDimensionIndex: <customDimensionIndex>,
      cookieDuration: 14, // set to 14 days
      contentExperimentId: <contentExperimentId>,
      cohorts: {
        original: { callback: function() {}, variantId: 0},
        next: { html: 'Next', variantId: 1 },
        continue: { html: 'Continue', variantId: 2 }
      }
    });
```
### Sample outcome
screen shot 2016-07-13 at 15 33 38

### DIFF
--- a/docs/javascript.md
+++ b/docs/javascript.md
@@ -178,6 +178,20 @@ var test = new GOVUK.MultivariateTest({
 });
 ```
 
+If you want to set the cookie expiration then you can optionally set cookieDuration as follows:
+
+```javascript
+var test = new GOVUK.MultivariateTest({
+  name: 'car_tax_button_text',
+  cookieDuration: 14,
+  cohorts: {
+    pay_your_car_tax: {weight: 25, callback: function() { ... }}, // 25%
+    give_us_money:    {weight: 75, callback: function() { ... }}  // 75%
+  }
+});
+```
+Here, it is set to expire in 14 days time. if this option is not set the default cookie expiration (30 days) take effect.
+
 If you have a complex test, it may be worth extending MultivariateTest with
 your own. Callbacks can be strings which will call a method of that name
 on the current object.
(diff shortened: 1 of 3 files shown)


dsingleton:
Thanks @ikennaokpala, being able to set a duration for the experiment is a really useful feature, and was needed recently in some other work. So 👍  to the feature idea.

Could you [add a test](https://github.com/alphagov/govuk_frontend_toolkit/blob/master/spec/unit/multivariate-test.spec.js) for the new behaviour? There's some prior art for checking cookie values, so I think it should be easy. 

One small, non-blocking, point: Is days is the best unit for the duration? [Recently](https://github.com/alphagov/frontend/pull/974) we needed a duration in minutes, rather than days,  which was possible with *Maths*, but makes me wonder if we should use seconds, would that be more consistent? 

If there's no consensus on that, i'd be more than happy to merge with days, but worth discussing until theres tests.

dsingleton:
👍  i'm fine with the duration being in days, it's a sensible default and if someone needs to set a lower duration then the caller can calculate a fraction value.

PR #293, opened 2016-07-13
Remove external link styles

robinwhittleton:
This removes the styles for external links and their associated icons. It also removes the documentation. Will need discussion before a merge happens.


joelanman:
I think this issue has two parts:
1. Is there a user need to know that a link takes them away from gov.uk?
2. If there is, what is a clear way to meet that need?

I'm not aware of any clear need to know that a link goes away from gov.uk.

In terms of the icon itself, it's ambiguous. The box and arrow icon is commonly used to mean 3 things:
 - [External link](https://www.google.co.uk/search?q=external+link+icon&source=lnms&tbm=isch&sa=X&ved=0ahUKEwjqgISTkvHNAhXFWywKHe9zDyYQ_AUICCgB&biw=1306&bih=737)
 - [Open new window](https://www.google.co.uk/search?q=new+window+icon&oq=new+window+icon&aqs=chrome.0.69i59j69i57j69i60l2j69i64j69i60.1806j0j8&sourceid=chrome&ie=UTF-8)
 - [Share](https://www.google.co.uk/search?q=share+icon&source=lnms&tbm=isch&sa=X&ved=0ahUKEwif6qjglPHNAhXKLMAKHcs1Bd4Q_AUICCgB&biw=1306&bih=737)

In light of this, the icon is adding noise to the page (cognitive load), and potentially even misleading people as to what will happen. If there really is a need, we need to find a clearer way to meet it.

tombye:
I'd add a few points to that.

1. do users know they are 'on' www.gov.uk, what does this mean to them and what does leaving it mean?
2. what information is useful when deciding to click an external link or not and does the visual design provide it?

timpaul:
We should use Hackpad for this discussion: https://designpatterns.hackpad.com/Links-CgBO9L4bF4a

@joelanman - would you take the lead on drafting some guidance? FWIW I think we should at least remove them as a default style, but we need a clear position on linking to external sites to do that.

robinwhittleton:
If we remove them as a default style then we’ll want to do the work in static and elements rather than here.

(also the PR in elements [does that update](https://github.com/alphagov/govuk_elements/pull/274/files#diff-64c472cbe8c69ec41b40abcd0c771415L9) @timpaul)

dsingleton:
I think the Content Team will have some good opinions on use and benefit of external links, a lot of external link usage is going to be in managed content.

whitingx:
Transcript of discussion around this ⤴ topic from the _Design Systems_ Slack community (http://designsystems.herokuapp.com/) - if useful ツ

> dsingleton [10:17 AM]  
> Does anyone have strong opinions, or research/evidence, on highlighting external links (eg, with icons)?
> 
> [10:18]  
> www.gov.uk uses them by default at the moment, but we’re looking at removing, https://github.com/alphagov/govuk_frontend_toolkit/pull/293 - any thoughts?
>  GitHub
> Remove external link styles [DISCUSS BEFORE MERGE] by robinwhittleton · Pull Request #293 · alphagov/govuk_frontend_toolkit · GitHub
> This removes the styles for external links and their associated icons. It also removes the documentation. Will need discussion before a merge happens. 
> 
> whitingx [10:37 AM]  
> @dsingleton: our DS currently uses an icon :arrow_upper_right: to differentiate external from internal links - but there are some good points in that PR discussion in regard to how much difference this actually makes to users :thinking_face:
> 
> ​_Wikipedia_​, who do use an icon to differentiate internal/external links, have some useful guidance/discussion in this area - https://en.wikipedia.org/wiki/Wikipedia%3aExternal_links - but think this is mainly preference/design rather then research based.
> 
> Would be interested to know any research you do find in this area ツ
> 
> dsingleton [10:39 AM]  
> We’ve not done research on external links themselves, but they’ve come up a few times in other research, and ​_generally_​ people don’t seem to know/care
> 
> [10:40]  
> A lot of times users don’t even know they’ve left GOV.UK and gone to a different site, even once they’re on it
> 
> [10:41]  
> That wikipedia link is interesting, thanks!
> 
> whitingx [10:42 AM]  
> > generally_ people don’t seem to know/care - @dsingleton
> 
> Could well imaging this would be the case around any research done - would think if the user clicked an unmarked external link they would simply click the :back: button to return to the origin site :thinking_face: - not sure what difference having the icon would actually make here.
> 
> And the point made in the PR;
> 
> ```In terms of the icon itself, it's ambiguous. The box and arrow icon is commonly used to mean 3 things:
> 
> External link
> Open new window
> Share```
> 
> is a good point that the addition of an icon is just additional visual clutter
> 
> dsingleton [10:43 AM]  
> I think an argument for is that it should indicate reduced trust, a less authoritative source
> 
> [10:43]  
> but there’s not a lot of evidence that is true ¯\_(ツ)_/¯
> 
> whitingx [10:49 AM]  
> Looking at how we currently structure external links—wondering if adding in the external/base URL might be a better way to identify this change of location to the user :thinking_face:
> 
> Move from;
> 
> More info can be found on `External Link Title ↗️` if interested.
> 
> to;
> 
> More info can be found on `External Link Title [https://en.wikipedia.org]` if interested.
> 
> :thinking_face: (edited)
> 
> robinwhittleton_gds [10:50 AM]  
> Do people understand URLs enough for that to be useful?
> 
> whitingx [10:50 AM]  
> @robinwhittleton_gds: probably not ツ
> 
> whitingx [10:56 AM]  
> A really interesting discussion point this @dsingleton / @robinwhittleton_gds
> 
> Quite a common pattern to indicate/differentiate internal/external links but no real evidence/research to back the pattern up :smile:
> 
> dsingleton [11:08 AM]  
> Yeah, my gut is that it’s more of a content writing problem than technical
> 
> gemmaleigh [11:13 AM]  
> joined #general
> 
> whitingx [11:15 AM]  
> Considering this from a semantic level—I’m now thinking that, from an end-user perspective (in our own DS ツ ), a `hyperlink` should just be a `hyperlink` - in its most basic form, whether it links to an internal page, an external page, a PDF, etc. it’s performing the same function of navigating the user from `A` > `B`
> 
> As such, assuming this basic ​_function_​ isn’t different, why should the styling be :thinking_face: (edited)
> 
> ed_gds [11:20 AM]  
> There are some specialist cases where I’ve found them helpful (though that might be the new window icon) - for example, on google analytics how it will distinguish links that go to the ​_real_​ service from links that go to pages ​_about_​ the service.
> 
> dsingleton [11:20 AM]  
> Thats a good example, and those are new window right?
> 
> ed_gds [11:21 AM]  
> I ​*think*​ so.
> 
> dsingleton [11:22 AM]  
> GA is more app-like, so it’s more like “go to the web/exit app” (reckon)
> 
> ed_gds [11:23 AM]  
> After trying it, the link goes to the page about the page, and the ​_icon_​ goes to the real url
> 
> [11:23]  
> :shrug
> 
> whitingx [11:29 AM]  
> Also wonder about accessibility related considerations for external link styling - not sure if screen-readers would explicitly identify the :arrow_upper_right: icon - wonder if some kind on inline link explanation e.g.
> ```External Link Text [https://en.wikipedia.org]```
> would be of benefit here :thinking_face: (edited)
> 
> [11:33]  
> … some useful reference discussion on ​_UX Stackexchange_​ in this area;
> 
> http://ux.stackexchange.com/questions/31498/visual-differentiation-between-in-site-links-to-outbound-links
> 
> ```Do not distinguish between in-site and out-site links.
> 
> Why? People don't care. They are using the internet to find information and will gladly follow the links they feel will get them to where they want to go.
> 
> In my experience (admittedly, this is by no means a scientific based thought) the belief that out-bound links should be treated differently stems from clients feeling that their site is more important than their customers feel that it is.
> 
> You often hear arguments "we need to launch external links in a pop-up so that users can come back to our site". I find that an arrogant belief that the users want to come back. They may. And if they do, they can certainly use that back button. But odds are they are really just focused on getting the information they are looking for. And if your site leads them to that information in an unobtrusive as possible way, that's a good user experience.
> ```
>  ux.stackexchange.com
> Visual differentiation between in-site links to outbound links
> I'm trying to solve the differentiation between in-site links vs. outbound links. Currently I'm using for all links the blue from my palette and underline on hover. Due to the nature of the site, t... 
>  
> new messages
> kim.boe [11:50 AM]  
> obviously anecdotal, but in the product I’m currently working on (a web-app, not website) users have to jump to external domain for certain operations requiring separate authentication… during user-testing, all users struggled to A: realize they had gone to external domain and B: navigate back using browser controls… added external link icons and in the next round of tests we had no issues.
> 
> On websites, I suspect there wouldn’t be too much of a difference… users are expecting to jump around. For web-apps and logged-in experiences, I think it is probably helpful with some distinction, as the user can become very disoriented if they are sent to a new context without some warning...

vipickering:
If it's helpful. Here is a good page with a lot of external links, that removing them would not necessarily be clear the user would be navigating off the website.

https://www.gov.uk/help/browsers

robinwhittleton:
Another example where they aren’t working well inside the design: https://www.gov.uk/government/organisations/rail-accident-investigation-branch#corporate-info has a custom link colour that the icon isn’t matching.

joelanman:
can we merge this? we have no clear user need for them, and they cause formatting issues

image

edwardhorsford:
I don't think they _cause_ formatting issues, but we have do have bugs relating to them.

If we don't merge this, we should fix the bugs with the display of the icons.

dsingleton:
> If we don't merge this, we should fix the bugs with the display of the icons.

Fixing the display bugs is always going to be a case by case thing - or putting in extra complexity to make it work automatically.

From a technical point or view, i'm pro-removing the icons, as it simplifies how we do text in general, and requires less special casing.

To make some progress with this change, how we  start by removing it as a default in [`govuk_template`](https://github.com/alphagov/govuk_template/blob/b6d4e882d195ed5093beb9fa1396438773c17de3/source/assets/stylesheets/_basic.scss#L97) - with a change note explaining how to bring it back if you want it?

It's a much less contentious first change - so doesn't risk people avoiding upgrades - and we can use it to gauge use of external links before removing it entirely?

robinwhittleton:
I’d be happy with that for the time being. Can treat it as a deprecation warning and let it bake for a few months.

Having said that, if design do decide that we just want to drop them completely that’s also fine by me.

edwardhorsford:
If we do depreciate / drop, it would be good if the changelogs put a call out for any services to let us know if this negatively effects them, or if they have any research showing the icons are needed or used.

esr360:
+1 for removing these -  the provided code is not pretty and we always end up using custom implementation of the external link icon anyway.

robinwhittleton:
I’ve just checked the HTML5.1 spec and `external` isn’t listed as a valid value for `rel`: https://www.w3.org/TR/html51/links.html#sec-link-types

nickcolley:
Related
- http://microformats.org/wiki/existing-rel-values -> http://microformats.org/wiki/rel-external 
- https://blog.whatwg.org/the-road-to-html-5-link-relations#rel-external

robinwhittleton:
A post on digital-service-designers today showed the use of this icon to mean ‘opens in new tab’ (they’d applied it to links in the footer so as not to break flow). Shows they’re sometimes misinterpreted even by designers.

esr360:
I've always seen this as a "new tab icon". As far as the user is concerned, they probably care more about whether the link they are clicking will take them to a new tab or break their current flow.

robinwhittleton:
[This thread on digital-service-design](https://groups.google.com/a/digital.cabinet-office.gov.uk/d/topic/digital-service-designers/YgZxuu51nKI/discussion) was mostly people +1ing the proposal to remove.

When we went to v2.0.0 of elements a month ago we added a note to the changelog deprecating external links and asking people to get in touch with any findings from user research they’ve taken. As far as I know we haven’t received anything from that.

dsingleton:
👏

PR #294, opened 2016-07-13
Allow use of multiple `customDimensionIndex` option

ikennaokpala:
## Description 
In this commit I have added code to allow more that one `customDimensionIndex` to be used during AB test.

Custom dimensions in Google Analytics can only have one of three scopes (hit, session, or user).
We need to explore how we measure the performance of A/B tests, and want to compare test results for
session-scoped custom dimensions (eg: index 222) against hit-scoped ones (eg: index 223).

## Sample code

```js

//= require govuk/multivariate-test

$(function(){
  if(isExpectedPath("overseas-passports")) {
    new GOVUK.MultivariateTest({
      el: '.get-started a',
      name: 'startButton_201607',
      customDimensionIndex: 13,
      contentExperimentId: '<customDimensionIndex>',
      cohorts: {
        original: { callback: function() {}, variantId: 0 },
        getApplicationInfo: { html: 'Get application information', variantId: 1 },
        next: { html: 'Next', variantId: 2 }
      }
    });
  }

  if(isExpectedPath("calculate-your-child-maintenance")) {
    new GOVUK.MultivariateTest({
      el: '.get-started a',
      name: 'startButton_201607',
      customDimensionIndex: [13, 14],
      contentExperimentId: '<customDimensionIndex>',
      cohorts: {
        original: { callback: function() {}, variantId: 0 },
        calculate: { html: 'Calculate', variantId: 1 },
        estimateChildMaintenance: { html: 'Estimate your child maintenance', variantId: 2 }
      }
    });
  }

  if(isExpectedPath("marriage-abroad")) {
    new GOVUK.MultivariateTest({
      el: '.get-started a',
      name: 'startButton_201607',
      customDimensionIndex: 14,
      contentExperimentId: '<customDimensionIndex>',
      cohorts: {
        original: { callback: function() {}, variantId: 0},
        next: { html: 'Next', variantId: 1 },
        continue: { html: 'Continue', variantId: 2 }
      }
    });
  }

  function isExpectedPath(slug) {
    var currentPath = window.location.pathname.split("/").join("");
    return currentPath == slug
  }
});

```

#### Expected result

* customDimensionIndex is set to 14 only

screen shot 2016-07-13 at 16 14 07

#### Expected result

* customDimensionIndex is set to 13 and 14 (hit and session)

screen shot 2016-07-13 at 16 15 07

### DIFF
--- a/javascripts/govuk/multivariate-test.js
+++ b/javascripts/govuk/multivariate-test.js
@@ -76,12 +76,21 @@
   };
 
   MultivariateTest.prototype.setCustomVar = function(cohort) {
-    if (this.customDimensionIndex) {
-      GOVUK.analytics.setDimension(
-        this.customDimensionIndex,
-        this.cookieName() + "__" + cohort
-      );
+    if (this.customDimensionIndex &&
+      this.customDimensionIndex.constructor === Array) {
+      for (var index = 0; index < this.customDimensionIndex.length; index++) {
+        this.setDimension(cohort, this.customDimensionIndex[index])
+      }
+    } else if (this.customDimensionIndex) {
+      this.setDimension(cohort, this.customDimensionIndex)
     }
+  };
+
+  MultivariateTest.prototype.setDimension = function(cohort, dimension) {
+    GOVUK.analytics.setDimension(
+      dimension,
+      this.cookieName() + "__" + cohort
+    );
   };
 
   MultivariateTest.prototype.setUpContentExperiment = function(cohort) {
--- a/docs/javascript.md
+++ b/docs/javascript.md
@@ -264,6 +264,24 @@ var test = new GOVUK.MultivariateTest({
 
 `customDimensionIndex` is the index of the custom variable in Google Analytics. GA only gives 50 integer slots to each account, and it is important that a unique integer is assigned to each test. Current contact for assigning a custom var slot for GOV.UK is: Tim Leighton-Boyce <tim.leighton-boyce@digital.cabinet-office.gov.uk>
 
+For certain types of tests where the custom dimensions in Google Analytics can only have one of three scopes (hit, session, or user). Measuring performance of A/B tests, and want to compare test results for session-scoped custom dimensions (eg: index 222) against hit-scoped ones (eg: index 223) may become important.
+
+Instead of supplying an integer (like the above example), `customDimensionIndex` can also accept an array of dimension indexes ([222, 223]), see below for more.
+
+Make sure to check GA debugger that these values are being sent before deploying.
+
+```js
+var test = new GOVUK.MultivariateTest({
+  name: 'car_tax_button_text',
+  customDimensionIndex: [222, 223],
+  cohorts: {
+    pay_your_car_tax: {weight: 25},
+    give_us_money: {weight: 50}
+  }
+});
+```
+
+
 ## Primary Links
 
 `GOVUK.PrimaryList` hides elements in a list which don't have a supplied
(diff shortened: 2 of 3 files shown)


nickcolley:
There's some usage of `forEach` and `Array.isArray` which is ES5 and won't be supported in legacy browsers.

We could consider adding a polyfill, rewriting in ES3...

javascripts/govuk/multivariate-test.js
```diff
@@ -75,14 +75,23 @@
...
-    if (this.customDimensionIndex) {
-      GOVUK.analytics.setDimension(
-        this.customDimensionIndex,
-        this.cookieName() + "__" + cohort
-      );
+    if (this.customDimensionIndex && Array.isArray(this.customDimensionIndex)) {
+      var _this = this;
+      this.customDimensionIndex.forEach(function(dimensionIndex) {
+        _this.setDimension(cohort, dimensionIndex)
+      });
```
nickcolley: Could consider using `function(){}.bind(this)` to give this closure the right context.

Then you don't need to assign a `_this` variable.

ikennaokpala: @nickcolley I have updated the code to bind as you suggested..

ikennaokpala:
@nickcolley I am not very conversant with the difference between ES3/ES5. 

Is there existing or a way to setup and test these within isolated environments for each. 

it would be nice.. that way I can validate this.

Other options that may suffice 

```js
if (this.customDimensionIndex &&
  this.customDimensionIndex.constructor == Array) {
```

```js
if (this.customDimensionIndex &&
  this.customDimensionIndex instanceof Array) {
```

nickcolley:
I think the best you can do to test in older legacy browsers and see if the code executes, since ES5 is not implemented 100% in different browsers etc.

I think we should avoid writing ES3 code where there are more expressive and maintainable ways to write them in ES5/6 (like you have in your current pull request).

So with that in mind we could consider:

First is polyfilling newer ES5 features, we have done this with .bind() here https://github.com/alphagov/govuk_frontend_toolkit/blob/master/javascripts/vendor/polyfills/bind.js

So we could add features as we go or consider something like https://github.com/es-shims/es5-shim.

Downside to this is we're adding more for our users to download.

We could use jQuery like a utility library but that makes the code dependant on jQuery which is not ideal.

nickcolley:
'Offline' discussion on internal Slack we came to the conclusion that we'll go with ES3 and discuss polyfilling etc in a broader scale later.

ikennaokpala:
@nickcolley Yeah I have made those changes..

nickcolley:
@ikennaokpala just noting for anyone who was not in Slack 👍

javascripts/govuk/multivariate-test.js
```diff
@@ -76,14 +76,23 @@
   };
 
   MultivariateTest.prototype.setCustomVar = function(cohort) {
-    if (this.customDimensionIndex) {
-      GOVUK.analytics.setDimension(
-        this.customDimensionIndex,
-        this.cookieName() + "__" + cohort
-      );
+    if (this.customDimensionIndex &&
+      this.customDimensionIndex.constructor == Array) {
```
nickcolley: Sorry for the nit picking but we could do `===` instead of `==` here which.

(Once we have proper linting I won't have to make these annoying comments sorry :P)

dsingleton:
LGTM, if no one else has any feedback i'll merge in a few minutes :)

PR #295, opened 2016-07-21
Fix this. reference in selection-buttons.js

### LINKED ISSUE #267: Possible bug in selection-buttons.js
Hello!

We've just started using eslint, and it's flagged up a possible issue in `selection-buttons.js` - I think it's correct.

The reported error is that `$elms` https://github.com/alphagov/govuk_frontend_toolkit/blob/master/javascripts/govuk/selection-buttons.js#L9 is not used.

It's assigned a little further down https://github.com/alphagov/govuk_frontend_toolkit/blob/master/javascripts/govuk/selection-buttons.js#L19 - but then this variable is never used.

I think it should be `this.$elms` if I'm following correctly.

gemmaleigh:
Moving this commit over from the govuk_prototype_kit repo.

This commit was made by @tvararu:
https://github.com/alphagov/govuk_prototype_kit/pull/223/commits/5caea24
efa4302572636a91288d58da0b566376f

This fixes #267.

### DIFF
--- a/javascripts/govuk/selection-buttons.js
+++ b/javascripts/govuk/selection-buttons.js
@@ -5,7 +5,6 @@
   var GOVUK = global.GOVUK || {};
 
   var SelectionButtons = function (elmsOrSelector, opts) {
-    var $elms;
 
     this.selectedClass = 'selected';
     this.focusedClass = 'focused';
@@ -15,7 +14,7 @@
       }.bind(this));
     }
     if (typeof elmsOrSelector === 'string') {
-      $elms = $(elmsOrSelector);
+      this.$elms = $(elmsOrSelector);
       this.selector = elmsOrSelector;
       this.setInitialState($(this.selector));
     } else if (elmsOrSelector !== undefined) {


gemmaleigh:
Closing while I check the tests...

gemmaleigh:
The tests fail when this change is made:

```
When GOVUK.SelectionButtons is called with a selector and then the page content is replaced
     When that selector matches radio inputs
       If one of those radios is clicked
         X Should mark that radio with the selected class
           Expected false to be true. (1)
         ✓ Should remove the selected class from all other radios
       If one of those radios is focused
         X Should add the focused class to the radio
           Expected false to be true. (1)
         X Should add a custom focused class to the radio if sent in as an option
           Expected false to be true. (1)
         X Should remove the focused class from a radio when it loses focus
           Expected false to be true. (1)
     When that selector matches checkbox inputs
       If one of those checkboxes is clicked
         X Should add the selected class to the checkbox
           Expected false to be true. (1)
       If one of those checkboxes is focused
         X Should add the focused class to the checkbox
           Expected false to be true. (1)
         - Should add a custom focused class to the checkbox if sent in as an option.         X Should add a custom focused class to the checkbox if sent in as an option
           Expected false to be true. (1)
         X Should remove the focused class from the checkbox when it loses focus
           Expected false to be true. (1)
   GOVUK.SelectionButtons.prototype.destroy
     - Should remove the events bound to the jQuery-wrapped elements sent into GOVUK.     ✓ Should remove the events bound to the jQuery-wrapped elements sent into GOVUK.SelectionButtons
     - Should remove the events bound to the document for the selector was sent into      X Should remove the events bound to the document for the selector was sent into GOVUK.SelectionButtons
       Expected false not to be false. (1)
       Expected false not to be false. (2)
       Expected Function to equal false. (3)
       Expected Function to equal false. (4)
```

PR #297, opened 2016-07-26
Feature/anchor buttons

paulmsmith:
As discussed in '[remove button role from start link](https://github.com/alphagov/govuk_elements/pull/272)' here is a pull request to shim anchors with ```role="button"``` to be 'clicked' via the space key.

I've not written tests yet, assuming the general style of this is fine then I'll do that. My thoughts were that this shim should be configurable but I've heard opinion that for a 'shim' that is not always a good thing. Thoughts, Suggestions welcome. :)

### DIFF
--- a/javascripts/govuk/anchor-buttons.js
+++ b/javascripts/govuk/anchor-buttons.js
@@ -1,0 +1,55 @@
+(function(global, customConfig) {
+  "use strict"; 
+       
+  var $ = global.jQuery;
+  var GOVUK = global.GOVUK || {};  
+
+  GOVUK.anchorButtons = {
+    
+    // default configuration that can be overridden by passing object as second parameter to module
+    config: $.extend({
+      // setting to true will disable the init function upon execution
+      disableInit: false,
+      // the target element(s) to attach the shim event to
+      selector: 'a[role="button"]',
+      // array of keys to match against upon the keyup event
+      keycodes: [
+        32 // spacekey
+      ],
+    }, customConfig),
+    
+    // event behaviour (not a typical anonymous function for resuse if needed)
+    triggerClickOnTarget: function triggerClickOnTarget(event) {
+      var code = event.charCode || event.keyCode; 
+      // if the keyCode/charCode from this event is in the keycodes array then 
+      if ($.inArray(code, this.config.keycodes) !== -1) {
+        event.preventDefault();
+        // trigger the target's click event
+        $(event.target).trigger("click");
+      }
+    },
+    
+    // init function (so it can be executed again if needed)
+    init: function init() { 
+      var $elms = $(this.config.selector);
+      // if found elements that match the selector in config (or customConfig) then
+      if($elms.length > 0) {
+        // iterate them giving access to current scope
+        $elms.each(function(index, elm){
+          // attach triggerClickOnTarget with current scope to the keyup event
+          $(elm).on('keyup', this.triggerClickOnTarget.bind(this));
+        }.bind(this));
+      }
+    }
+
+  };
+
+  // if disbaleInit is not true then run the init method
+  if(!GOVUK.anchorButtons.config.disableInit) {
+    GOVUK.anchorButtons.init(); 
+  }
+  
+  // hand back to global
+  global.GOVUK = GOVUK;
+      
+})(window);
--- a/javascripts/govuk/anchor-buttons.js
+++ b/javascripts/govuk/anchor-buttons.js
@@ -1,4 +1,14 @@
-(function(global, settings) {
+// javascript 'shim' to trigger the click event of element(s) 
+// when the space key is pressed.
+// 
+// usage instructions:
+// GOVUK.anchorButtons.init();
+// 
+// If you want to customise the shim you can pass in a custom configuration
+// object with your own selector for the target elements and addional keyup 
+// codes if there becomes a need to do so. For example:
+// GOVUK.anchorButtons.init({ selector: '[role="button"]' }); 
+(function(global) {
   "use strict"; 
        
   var $ = global.jQuery;
@@ -7,16 +17,14 @@
   GOVUK.anchorButtons = {
     
     // default configuration that can be overridden by passing object as second parameter to module
-    config: $.extend({
-      // setting to true will disable the init function upon execution
-      disableInit: false,
+    config: {
       // the target element(s) to attach the shim event to
       selector: 'a[role="button"]',
       // array of keys to match against upon the keyup event
       keycodes: [
         32 // spacekey
       ],
-    }, settings),
+    },
     
     // event behaviour (not a typical anonymous function for resuse if needed)
     triggerClickOnTarget: function triggerClickOnTarget(event) {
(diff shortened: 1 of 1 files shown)


javascripts/govuk/anchor-buttons.js
```diff
@@ -0,0 +1,55 @@
...
+        $elms.each(function(index, elm){
+          // attach triggerClickOnTarget with current scope to the keyup event
+          $(elm).on('keyup', this.triggerClickOnTarget.bind(this));
+        }.bind(this));
+      }
+    }
+
+  };
+
+  // if disbaleInit is not true then run the init method
```
alextea: typo ;)

javascripts/govuk/anchor-buttons.js
```diff
@@ -0,0 +1,55 @@
...
+      if ($.inArray(code, this.config.keycodes) !== -1) {
+        event.preventDefault();
+        // trigger the target's click event
+        $(event.target).trigger("click");
+      }
+    },
+    
+    // init function (so it can be executed again if needed)
+    init: function init() { 
+      var $elms = $(this.config.selector);
```
nickcolley: Could we attach a listener to the document rather than one for each button?

```js
$(document).on('keyup', this.config.selector, this.triggerClickOnTarget.bind(this));
```

My thought is that this way we only have one delegated listener rather than potentially a few?

javascripts/govuk/anchor-buttons.js
```diff
@@ -0,0 +1,55 @@
...
+  var GOVUK = global.GOVUK || {};  
+
+  GOVUK.anchorButtons = {
+    
+    // default configuration that can be overridden by passing object as second parameter to module
+    config: $.extend({
+      // setting to true will disable the init function upon execution
+      disableInit: false,
+      // the target element(s) to attach the shim event to
+      selector: 'a[role="button"]',
```
nickcolley: Can we add this role to other elements such as spans?

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_button_role

Maybe the selector should be just `[role="button"]`.

nickcolley:
Thanks a lot for getting this together Paul 👍 

One thing that worries me with this is it could break the behaviour of scrolling the page when pressing space.

It seems as though if a user is focused on a link it's okay to block that behaviour but I'm not entirely sure of all the edge cases so thought I'd comment in the hope that someone else can think about this too.

javascripts/govuk/anchor-buttons.js
```diff
@@ -0,0 +1,55 @@
...
+      keycodes: [
+        32 // spacekey
+      ],
+    }, settings),
+    
+    // event behaviour (not a typical anonymous function for resuse if needed)
+    triggerClickOnTarget: function triggerClickOnTarget(event) {
+      var code = event.charCode || event.keyCode; 
+      // if the keyCode/charCode from this event is in the keycodes array then 
+      if ($.inArray(code, this.config.keycodes) !== -1) {
```
nickcolley: A nit (sorry) but I'm wondering, if we're only ever going to be shimming the spacekey, could we just do a boring

```js
code === 32
```

here instead of an array comparison, not super important (sorry again :P)

javascripts/govuk/anchor-buttons.js
```diff
@@ -0,0 +1,55 @@
+(function(global, customConfig) {
```
robinwhittleton: Can we add a comment block at the top explaining what the file does? E.g. https://github.com/alphagov/govuk_frontend_toolkit/blob/master/javascripts/govuk/stop-scrolling-at-footer.js

javascripts/govuk/anchor-buttons.js
```diff
@@ -0,0 +1,55 @@
+(function(global, settings) {
```
robinwhittleton: Can we add a comment block at the top explaining what the file does? E.g. https://github.com/alphagov/govuk_frontend_toolkit/blob/master/javascripts/govuk/stop-scrolling-at-footer.js

robinwhittleton:
This adds a file, but won’t roll it out across GOV.UK. I guess we want to add it to [govuk_template.js](https://github.com/alphagov/govuk_template/blob/master/source/assets/javascripts/govuk-template.js), intialise it in [govuk_template’s core.js](https://github.com/alphagov/govuk_template/blob/master/source/assets/javascripts/core.js), then push a new version and update [static’s Gemfile](https://github.com/alphagov/static/blob/master/Gemfile) to use that.

javascripts/govuk/anchor-buttons.js
paulmsmith: Yep, good point I think that's probably better. 👍  Will do that.

javascripts/govuk/anchor-buttons.js
paulmsmith: I wasn't sure about that one. Do we always want space to trigger a click on something with ```role="button"``` ? @LJWatson @aduggin to advise perhaps?

javascripts/govuk/anchor-buttons.js
paulmsmith: Again fair point. Originally there was talk of triggering on enter as well. Left it open to additional key-codes just in case more than anything. Happy to change if we feel that makes it more readable/explicit.

javascripts/govuk/anchor-buttons.js
paulmsmith: Sure..  I'll get that done and reply back when I've updated this PR.

javascripts/govuk/anchor-buttons.js
paulmsmith: Good spot! Will fix.

paulmsmith:
Hi all, I've made the updates after the suggestions. I have changed it now so that the event listener is on the document and I have also updated the module pattern so that it needs to be explicitly initiated.

javascripts/govuk/anchor-buttons.js
nickcolley: Generally for me I prefer to avoid anticipating future updates since more often than not they don't happen but don't think this blocks merging, so up to you. ✌️

paulmsmith:
@robinwhittleton I'm due to go away for two weeks now. I've removed the 'WIP' from the title and I think this can be merged whenever you and others feel is a good time.

gemmaleigh:
@paulmsmith thanks for this! We'll review shortly. Have a good break!

gemmaleigh:
I'd like to get @aduggin's opinion on this before we merge.

javascripts/govuk/anchor-buttons.js
```diff
@@ -0,0 +1,61 @@
...
+      // if the keyCode/charCode from this event is in the keycodes array then 
+      if ($.inArray(code, this.config.keycodes) !== -1) {
+        event.preventDefault();
+        // trigger the target's click event
+        $(event.target).trigger("click");
+      }
+    },
+
+    // By default this will find all anchors with role attribute set to 
+    // 'button' and will trigger their click event when the spaceky (32) is pressed.
```
gemmaleigh: mini-nit, typo in comment (spacekey)

javascripts/govuk/anchor-buttons.js
```diff
@@ -0,0 +1,61 @@
...
+        $(document).on('keyup', this.config.selector, this.triggerClickOnTarget.bind(this));
+      }    
+    }
+
+  };
+  
+  // hand back to global
+  global.GOVUK = GOVUK;
+      
+})(window);
```
gemmaleigh: Missing EOF.

aduggin:
This approach looks good to me

robinwhittleton:
Thanks. Will merge this now, address @gemmaleigh’s nits in a further PR.

PR #299, opened 2016-07-29
[Do not merge] Update the UK Trade Investment organisation color to reflect new brand

nickcolley:
As part of https://govuk.zendesk.com/agent/tickets/1385632

In terms of accessibility, the new colour is a darker pink/red so contrast should be better where it's used but I'm not so sure of the reaching changes of this so would appreciate that context :)

image


nickcolley:
Looks like we need to add a new organisation. So will put this on hold until monday..

PR #303, opened 2016-08-09
Remove unused $elms variable in selection-buttons.js

tvararu:
Previously: https://github.com/alphagov/govuk_frontend_toolkit/pull/295

Turns out the `this.$elms` write isn't necessary since it wasn't being done before anyway. It was being written to `$elms` but not being read from at any later point.

cc @gemmaleigh


nickcolley:
LGTM 👍

PR #305, opened 2016-08-10
Add GOVUK.ShowHideContent to toolkit, introduce tests

colinrotherham:
Hi @robinwhittleton,

I've seen the `ShowHideContent()` component used by various service teams (including the prototype kit) but it's often just pasted into `application.js` where required.

One use is to add progressive disclosure support (reveals a hidden panel) to question and answer forms. For example, answering "Yes" to a question may ask for more information.

It's not ideal to have it copy & pasted around without tests. Copied code will become stale.

For reference, I've added the original (untested) version here:
https://gist.github.com/colinrotherham/b9d53d0ea51bde85d8a1f26231038479

I've added tests and rewritten the module into the latest format for this pull request. 

I noticed @gemmaleigh and @fofr discussed it here:
https://github.com/alphagov/govuk_elements/pull/183

Can it live here in this repo instead? If not, please advise where it should go!

Thanks


colinrotherham:
One example that shows testing is needed was that "out of the box" the component doesn't add the necessary `aria-controls` and `aria-expanded` attributes.

These are only added when a radio or checkbox is clicked.

colinrotherham:
Here it is, untested, in the prototype kit:
https://github.com/alphagov/govuk_prototype_kit/blob/master/app/assets/javascripts/application.js

javascripts/govuk/show-hide-content.js
```diff
@@ -0,0 +1,150 @@
...
+    // Hide toggled content for control
+    function hideToggledContent($control, $content) {
+      $content = $content || getToggledContent($control);
+
+      // Hide content
+      if (!$content.hasClass('js-hidden')) {
+        $content.addClass('js-hidden');
+        $content.attr('aria-hidden', 'true');
+
+        // If the controlling input, update aria-expanded
```
tombye: If `$control` isn't a controlling input, it looks like the call to `getToggledContent` at the top of this function wouldn't work as its parent label wouldn't have a reference to the content id. If that's true, is it worth doing the check for `aria-controls` early and returning if it's not found.

docs/javascript.md
```diff
@@ -398,3 +398,41 @@ GOVUK.selectionButtons = function (elms, opts) {
...
+To apply this behaviour to elements with the above HTML pattern, call the `GOVUK.ShowHideContent` constructor with their inputs:
+
+```
+var showHideContent = new GOVUK.ShowHideContent();
+showHideContent.init();
+```
+
+This will bind all events to the document, meaning any changes to content (for example, by AJAX) will not effect behaviour.
+
+You can also call `GOVUK.ShowHideContent` with a selector (defaults to `$(document.body)`):
```
tombye: I found this a little confusing to read just because it doesn't say what the selector is for. Can we explain it's for the element the event is bound to?

javascripts/govuk/show-hide-content.js
```diff
@@ -0,0 +1,150 @@
...
+      // Set aria-controls and defaults
+      if ($content.length) {
+        $control.attr('aria-controls', $content.attr('id'));
+        $control.attr('aria-expanded', 'false');
+        $content.attr('aria-hidden', 'true');
+      }
+    }
+
+    // Return toggled content for control
+    function getToggledContent($control) {
```
tombye: This method of getting the content based on the parent label data attribute involves a bit of DOM traversal which only seems needed before the ARIA attributes are set. Afterwards, whenever you need the content, its id can be got from the `aria-controls` attribute on the control which seems a bit less work.

Would it be worth putting this block directly into the `init` method when `getToggledContent` is called and having this function just check `aria-controls`?

tombye:
Hi @colinrotherham, thanks for taking the time to raise this.

I've added a few comments to the code based on my understanding of it from reading through the JS. Apart from those on the docs, none of them are real blockers but I'd be interested in what you think.

javascripts/govuk/show-hide-content.js
```diff
@@ -0,0 +1,150 @@
...
+  var GOVUK = global.GOVUK || {};
+
+  function ShowHideContent() {
+    var self = this;
+
+    // Radio and Checkbox selectors
+    var selectors = {
+      namespace: 'ShowHideContent',
+      radio: '.block-label input[type="radio"]',
+      checkbox: '.block-label input[type="checkbox"]'
```
nickcolley: We might want make sure we're not using presentational classes for binding, this will make this module more flexible.

(https://github.com/alphagov/styleguides/blob/master/js.md#html-class-hooks)

For example

```html
<label class="block-label js-toggle-showhide" data-toggle-target="show-me">
  <input type="radio" name="enabled" value="yes" /> Yes
</label>

<div id="show-me" class="panel js-hidden">
  <p>Show/Hide content to be toggled</p>
</div>
```

colinrotherham: Hi @nickcolley, don't forget this isn't my module. I've only refactored, tested and migrated it here.

If you want to change it's behaviour or styling let's open another pull request. It's already in use elsewhere using this class so we'd have to think about backwards compatibility.

javascripts/govuk/show-hide-content.js
colinrotherham: Hi @tombye, whilst it might not be a controlling input (i.e. can't open a show/hide panel), clicks on other radio inputs with the same `name` attribute are still allowed to hide the show/hide panel.

The controlling input will be told to close the panel via the loop on line 81: https://github.com/alphagov/govuk_frontend_toolkit/pull/305/files/eccbfa8054f812321adadde10bd14be50d34491e#diff-5fcf67c331373114554826652b07e141R81

You're right though. In that, if only one radio in the loop has show/hide content, I was happy for the others to silently be skipped over and not do anything.

docs/javascript.md
```diff
@@ -398,3 +398,41 @@ GOVUK.selectionButtons = function (elms, opts) {
...
+    <label class="block-label">
+      <input type="radio" name="enabled" value="no" /> No
+    </label>
+
+    <div id="show-me" class="panel js-hidden">
+      <p>Show/Hide content to be toggled</p>
+    </div>
+
+When the input's `checked` attribute is set, the show/hide content's `.js-hidden` class is removed and ARIA attributes are added to enable it.
+
```
colinrotherham: @tombye Yeah, good point. I've changed this.

docs/javascript.md
colinrotherham: @tombye Definitely. I've added clarity here too.

javascripts/govuk/show-hide-content.js
colinrotherham: @nickcolley For example, I've spotted you're also lacking a `role="region"` on the show-hide div and aren't passing `.focus()` when the content is toggled open/closed.

Just mentioned this to @gemmaleigh.

If you think I'm the right person to progress with this I'm happy to help?

javascripts/govuk/show-hide-content.js
colinrotherham: @tombye I've just improved `getToggledContent` instead. I'd only want one method to have ownership of "get the content" so didn't want to have another version in the init methods.

The new version defaults to checking `aria-controls` for the ID.

tombye: Liking the new version, especially the `id` variable. Makes it nice and explicit what's going on 👍

javascripts/govuk/show-hide-content.js
gemmaleigh: Hi @colinrotherham, thanks for taking the time to add this JS and the tests and documentation to accompany it. We'll happily take it from here, rather than pester you with change requests. Thanks for your help!

javascripts/govuk/show-hide-content.js
tombye: I think I just found it odd reading that inputs without an 'aria-controls' attribute or a parent label with a `data-target` attribute were being sent to `getToggledContent` which assumes either of those.

At the moment this error is being absorbed by jQuery but it seems like this could be avoided by just filtering the radios down so only those controlling content had their content hidden.

docs/javascript.md
tombye: Thanks for this, reads a lot easier now.

javascripts/govuk/show-hide-content.js
colinrotherham: No worries @tombye, did you see the original code I linked to in my opening comments?

The new code sticks to the theme of the original which also looped all radios and also tried to find the show/hide content for them all too.

With a pull request I never want to step on anyone's toes. If you'd like to change this in the future it's all yours! 😊

tombye: I saw the code, thanks for that, it helped explain the flow nicely.

You're not stepping on anyone's toes. You put together a clean, well documented and tested piece of code on your own time for the benefit of the community. Please don't think any part of my review ignores the effort or thought that went in :)

I'm happy to put any suggested changes together into a separate pull request 👍

colinrotherham: Brill. Thanks @tombye.

I always imagine if I change too much it won't get merged in!

The goals I had were to:

1. Reduce fragmentation (copying between other services)
2. Switch to event delegation (reduces handler overhead on large forms)
3. Reduce duplicated code between radios and checkbox flows
4. Check for already-toggled content on first run (new feature)
5. Add test coverage

* Added here: https://github.com/alphagov/govuk_frontend_toolkit/pull/305/files/eccbfa8054f812321adadde10bd14be50d34491e#diff-5fcf67c331373114554826652b07e141R120

Lastly, if the ES5 `Function.prototype.bind` method was available, we could optimise this bit too:
https://github.com/alphagov/govuk_frontend_toolkit/pull/305/files#diff-5fcf67c331373114554826652b07e141R115

Are there any thoughts about adding the ES5 shims to bring `.bind()`, `.trim()`, `.forEach()`, `Object.keys()` etc to IE8 and other older browsers?

I can see my work has been branched off and progressed here: https://github.com/alphagov/govuk_frontend_toolkit/tree/add-govuk-show-hide-content-js

Feel free to close this pull request.

tombye: We include a `.bind()` shim in this repo which is used in the selection buttons code:

https://github.com/alphagov/govuk_frontend_toolkit/blob/6d55f22dfcc063724f28d37fc9f61cbf89c2ce7b/javascripts/govuk/selection-buttons.js#L14

@edds wrote a nice pull request explaining the merits of adding it here:

https://github.com/alphagov/govuk_frontend_toolkit/pull/120

I can see more shims being adopted if the pros and cons are explained as is done here. Adding `.bind()` definitely made the selection buttons code a lot clearer.

colinrotherham: Ah fab, sorry I couldn't see one at first glance. Thanks @tombye.

gemmaleigh:
I'm closing this in favour of #315. 

I'd like to get #315 merged as this will work as-is and is an improvement on what is currently in application.js in GOV.UK elements, anything else should be raised as a new PR.

Thanks everyone! @colinrotherham @tombye @nickcolley.

PR #312, opened 2016-08-19
Add missing anchor link in the README

gemmaleigh:
This links to the new docs for the shim-links-with-button-role.js

cc. @robinwhittleton.


robinwhittleton:
Oops, good catch.

PR #315, opened 2016-08-22
Add GOVUK.ShowHideContent JavaScript

gemmaleigh:
Add GOVUK.ShowHideContent to toolkit, this is copied from #305.
This PR is rebased against master and has  the conflicts from #305 resolved. 

A link is added to the README to the documentation.
Standard JS has been used to lint the JS.

cc. @tombye, @nickcolley.

### DIFF
--- a/javascripts/govuk/show-hide-content.js
+++ b/javascripts/govuk/show-hide-content.js
@@ -48,7 +48,6 @@
 
     // Show toggled content for control
     function showToggledContent ($control, $content) {
-
       // Show content
       if ($content.hasClass('js-hidden')) {
         $content.removeClass('js-hidden')
@@ -80,22 +79,23 @@
     // Handle radio show/hide
     function handleRadioContent ($control, $content) {
 
-      // All radios in this group
-      var selector = selectors.radio + '[name=' + escapeElementName($control.attr('name')) + ']'
+      // All radios in this group which control content
+      var selector = selectors.radio + '[name=' + escapeElementName($control.attr('name')) + '][aria-controls]'
       var $radios = $control.closest('form').find(selector)
 
-      // Hide radios in group
+      // Hide content for radios in group
       $radios.each(function () {
         hideToggledContent($(this))
       })
 
-      // Select radio button content
-      showToggledContent($control, $content)
+      // Select content for this control
+      if ($control.is('[aria-controls]')) {
+        showToggledContent($control, $content)
+      }
     }
 
     // Handle checkbox show/hide
     function handleCheckboxContent ($control, $content) {
-
       // Show checkbox content
       if ($control.is(':checked')) {
         showToggledContent($control, $content)
--- a/javascripts/govuk/show-hide-content.js
+++ b/javascripts/govuk/show-hide-content.js
@@ -10,8 +10,8 @@
     // Radio and Checkbox selectors
     var selectors = {
       namespace: 'ShowHideContent',
-      radio: '.block-label input[type="radio"]',
-      checkbox: '.block-label input[type="checkbox"]'
+      radio: '.block-label[data-target] input[type="radio"]',
+      checkbox: '.block-label[data-target] input[type="checkbox"]'
     }
 
     // Escape name attribute for use in DOM selector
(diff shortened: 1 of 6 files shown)


spec/unit/show-hide-content.spec.js
```diff
@@ -0,0 +1,245 @@
+describe('show-hide-content', function () {
+  'use strict'
+
+  beforeEach(function () {
+
+    // Sample markup
+    this.$content = $(
+
+      // Radio buttons (yes/no)
+      '<form class="js-toggle-showhide">' +
```
tombye: Is this `js-toggle-showhide` class needed? I can't see it used in the tests or the module code.

gemmaleigh: Good spot! Thanks @tombye, I've removed those.

javascripts/govuk/show-hide-content.js
```diff
@@ -80,22 +79,23 @@
     // Handle radio show/hide
     function handleRadioContent ($control, $content) {
 
-      // All radios in this group
-      var selector = selectors.radio + '[name=' + escapeElementName($control.attr('name')) + ']'
+      // All radios in this group which control content
+      var selector = selectors.radio + '[name=' + escapeElementName($control.attr('name')) + '][aria-controls]'
```
colinrotherham: Hi @tombye, @gemmaleigh.

This has a side effect unfortunately… Now *only* controlling radios can be used to show/hide content.

Picture a yes or no question and answer form, for example "Do you have a partner"?

Previously the "Yes" radio (controlling) could toggle the content open with further content, and the "No" radio in the same `name` group (not controlling) could toggle the content closed again.

With this latest change, how do you close it once it's open?

colinrotherham: example

tombye: I'm pretty sure the method you show/hide content for radio groups works with the change in that scenario:

If you click 'yes', all content controlled by the group would be hidden by the loop on line 87. The condition on line 92 will return true and so show the content controlled by the event target.

If you click 'no', all content controlled by the group would still be hidden by the loop on line 87. The condition on line 92 will return false so no content is shown.

I ran the tests after making this change and they were green so figured it was ok. I made a jsbin to play around with it which seems to work as expected: https://jsbin.com/nademaxaxo/edit?html,js,output

colinrotherham: Thanks @tombye, that's great.

Forgot I wrote a test to cover that very scenario!
https://github.com/alphagov/govuk_frontend_toolkit/pull/315/files#diff-1cb888219674ec4080d3a7d2bdca5117R183

Still works as intended then 👍

tombye:
This seems mostly there. I have one question for @gemmaleigh @robinwhittleton @nickcolley @tvararu. I'd like to change `data-target` to `data-controls` so it's obvious that it's copied across into an `aria-controls` attribute.

Be interested in your thoughts on this.

nickcolley:
@tombye I think if we want to keep the first release backwards compatible we should leave as is, otherwise there's a lot we could do to improve these and would be better thought about separately?

javascripts/govuk/show-hide-content.js
```diff
@@ -17,7 +17,7 @@
     // Escape name attribute for use in DOM selector
     function escapeElementName (str) {
       var result = str.replace('[', '\\[').replace(']', '\\]')
-      return (result)
```
nickcolley: We could just return the above expression, result isn't used anywhere else so just keep it simple..

tombye: Agreed and done.

tombye:
@gemmaleigh I'm done on my changes. It would be worth getting another pair of eyes on it before merging now we've both worked on it but it's 👍  by me.

gemmaleigh:
Hi @tombye, thanks for these. 

I've updated the JavaScript in a test branch for GOV.UK elements - so if anyone else would like to preview this working - [it can be seen here](http://govuk-elements-test.herokuapp.com/form-elements/example-radios-checkboxes/).

I'm keen that we don't change the `data-target` attribute for now, [it appears in quite a few places](https://github.com/search?l=html&q=org%3Aalphagov+data-target&type=Code).

Happy for this to be merged, then I'll bump the toolkit and update govuk_elements and the govuk_prototype kit.

PR #317, opened 2016-08-27
Mark selection buttons with a class that matches their input

robinwhittleton:
In govuk_elements we’d like to differentiate between selection buttons that contain a radio and those that contain a checkbox. This initialises selection buttons with a namespaced class that we can bind to.


tombye:
Looks good to me 👍

nickcolley:
Thanks @robinwhittleton 👍

PR #319, opened 2016-09-05
Close the example code block

gemmaleigh:
This PR adds the missing line to close the example block.

Screenshot before:

example-before

Screenshot after:

screenshot-after


nickcolley:
Good catch 👍

PR #321, opened 2016-09-05
Fix link to the show/hide content docs

gemmaleigh:
Using a `/` between the two words has meant they aren't separated by a hyphen, this PR fixes that - so the docs can be reached from the README.


nickcolley:
💥  nice thanks :)

PR #327, opened 2016-09-19
Remove grunt-shell dependency and use npm script

nickcolley:
I can't actually get `npm test` to work on my machine because of the ruby dependencies but this should work.

Should speed up `npm install` by a second or so..

```
--------------------------------------------------
| Dependency            | Time | Size   | # Deps |
--------------------------------------------------
| grunt-contrib-jasmine | 9.8s | 53 MB  | 111    |
| grunt                 | 3.4s | 7.6 MB | 30     |
| grunt-cli             | 1.8s | 1.3 MB | 11     |
| grunt-contrib-sass    | 1.7s | 186 KB | 13     |
| grunt-shell           | 1.6s | 74 KB  | 10     |
| jquery                | 1.3s | 1.3 MB | 0      |
| grunt-contrib-clean   | 1.1s | 23 KB  | 2      |
--------------------------------------------------
```

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -49,16 +49,8 @@ module.exports = function(grunt) {
           loadPath: [
             './stylesheets'
           ],
-          style: 'nested',
+          style: 'nested'
         }
-      },
-    },
-    shell: {
-      multiple: {
-        command: [
-          'bundle',
-          'bundle exec govuk-lint-sass stylesheets'
-        ].join('&&')
       }
     }
   });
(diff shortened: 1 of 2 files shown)


gemmaleigh:
I have tested this branch using:
`git clean -dfx`
`npm install`
`npm test`

Both tests run.

I also checked that the scss-linter is working by changing one of the colour values, the error was picked up. 

The npm error isn't very friendly, but the linter output is shown above.

nickcolley:
@robinwhittleton noted that if the script does fail there's a much more verbose error message that isn't as helpful for people to use.

One way to get around this could be to throw the `--silent`, I'll have a poke and see if this'll work.

robinwhittleton:
👍

PR #328, opened 2016-09-19
Remove `@media` width query from print styles

selfthinker:
This fixes that print styles using any of the typography mixins would have font sizes in `px` < 640px width and in `pt` after and sometimes the wrong line-height.
Different browser print views  also have different widths, making the print look less consistent between different browsers.

Please note, I did not test the changes as I don't know how test changes to the frontend toolkit locally.
This should also fix an outstanding issue on alphagov/govuk_template#237.

### DIFF
--- a/stylesheets/_typography.scss
+++ b/stylesheets/_typography.scss
@@ -35,18 +35,21 @@ $is-print: false !default;
       }
     }
   }
-  font-size: $font-size-640;
-  line-height: $line-height-640;
   font-weight: $font-weight;
   text-transform: none;
 
-  @include media(tablet) {
-    font-size: $font-size;
-    line-height: $line-height;
-  }
-
   @if $is-print {
     font-size: $font-size-print;
+    line-height: $line-height;
+
+  } @else {
+    font-size: $font-size-640;
+    line-height: $line-height-640;
+
+    @include media(tablet) {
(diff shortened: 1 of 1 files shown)


stylesheets/_typography.scss
```diff
@@ -40,13 +40,13 @@ $is-print: false !default;
   font-weight: $font-weight;
```
nickcolley: Looks like we'd still inherit the line-height from here?

Only conditionally having the media query feels arbitrary, maybe it'd be clearer to override the line-height for print?

selfthinker: The line-height isn't causing that much trouble. But you're right, it should be moved into the else case, together with the other font-size. And the normal font-size needs to be copied into the print style. I will adjust the commit accordingly.
I'm not sure I understand what you mean with "Only conditionally having the media query feels arbitrary". I would generally never have `@media` queries inside a print stylesheet. (Although I can think of legitimate reasons, I've never seen it used before.)

selfthinker: I just force-pushed the change.

nickcolley: Having less forks calms my simple mind 👍 

Will let someone else have a peek at this before merging..

robinwhittleton:
Happy with this.

PR #332, opened 2016-09-26
Allow custom options in GOVUK.analytics.trackPageview

colinrotherham:
Currently the `GOVUK.analytics.trackPageview()` method only supports the `transport` Google Analytics fieldsObject property, which puts a limit on its usefulness.

This limitation makes the following overrides unavailable:
`hitCallback`
`sessionControl`

This pull request allows any options object to be used.

For example, the current session can now be reset when returning to a specific page (perhaps when a kiosk session times out):
```javascript
GOVUK.analytics.trackPageview('/', 'Home', {
  sessionControl: 'start'
})
```


colinrotherham:
Hi @dsingleton, @nickcolley,

I've improved the way options can be set on `GOVUK.analytics.trackPageview()`, would you mind taking a look for me in this pull request?

The third `options`parameter I've extended isn't documented in **analytics.md** but it would be useful to others. Shall I add this or would you rather keep it as a hidden feature?

javascripts/govuk/analytics/google-analytics-universal-tracker.js
```diff
@@ -33,24 +34,25 @@
 
   // https://developers.google.com/analytics/devguides/collection/analyticsjs/pages
   GoogleAnalyticsUniversalTracker.prototype.trackPageview = function(path, title, options) {
-    var options = options || {};
+    var options;
```
robinwhittleton: Don’t think we need this line right? `options` is coming in as an argument and if it’s not supplied `if (typeof options === "object")` will just fail as `options` will equal `'undefined'`.

nickcolley:
Hey @colinrotherham 

Thanks for the PR! I had a look over the code it looks good, I think it'd be good to add the note to the docs if you have time.. 👍

javascripts/govuk/analytics/google-analytics-universal-tracker.js
colinrotherham: Good point. I saw it as a simplification, but missed that you were re-declaring `options`.

Let's tidy this up.

colinrotherham: The `trackEvent` method was doing the same.

Now fixed: https://github.com/alphagov/govuk_frontend_toolkit/pull/332/files#diff-8b93af583c1a7ca5e9c48a89ebce2b38L63

nickcolley: This would be caught when we add linting 👍

colinrotherham:
Thanks @nickcolley.

Spotted by @robinwhittleton were two places `options` was being redeclared (both `trackEvent()` and `trackPageview()` which are now sorted.

robinwhittleton:
So this looks conceptually fine to me, but I haven’t tested it on an actual site and I’m not going to have time to this week. @nickcolley, is this something you could do?

colinrotherham:
Hi @nickcolley,

Do you still plan to merge this in?

Thanks

nickcolley:
@colinrotherham I haven't had the time to test this sorry Colin will try to ASAP but hopefully someone else can review, will ping others too..

robinwhittleton:
Tested this with a few different inputs and it’s working as expected, so I’m going to merge. It would be good to get this tested with analytics people on GOV.UK before static is updated to include it.

PR #334, opened 2016-09-29
[Do note merge] [Request for comments] Lint codebase using standard

nickcolley:
I recognise this is a big diff, so just raising this to see what people think..

Best viewed commit by commit, any changes that could potentially change logic or were manually I've done in different commits - separate from the automated fixing.

Linting the codebase has removed unused variables, made us explicit about declaring globals and improved patterns like the following

```js
Foo(function (options) {
  var options = options || {}
}))
```

```js
Foo(function (options) {
  options = options || {}
}))
```

Finally I've silenced an error surrounding our use of 'new', this indicates an issue with our module pattern wherein we don't need to be using a class pattern for some of our code since there's no returned instance that is being used.

If this is the case that there are public methods exposed on an instance we should test those...


package.json
```diff
@@ -5,13 +5,14 @@
   "devDependencies": {
     "grunt": "~0.4.2",
     "grunt-cli": "0.1.11",
-    "grunt-contrib-clean":"~0.6.0",
+    "grunt-contrib-clean": "~0.6.0",
     "grunt-contrib-jasmine": "^1.0.0",
     "grunt-contrib-sass": "0.7.4",
-    "jquery": "~1.11.3"
+    "jquery": "~1.11.3",
+    "standard": "^8.2.0"
```
tvararu: NB: Standard 8 was launched this month but `govuk_prototype_kit` for ex is still on Standard 7.

Updating is not really a priority IMO. (the only important changes are for JSX I think which we don't use)

tvararu:
Can you add semicolons at the start of IIFE's as part of this? Will solve #338.

PR #329, opened 2016-09-20
Amend GOVUK.stickAtTopWhenScrolling to "unstick" for smaller screens

gemmaleigh:
For smaller screens (<768px) ensure that the element which was previously "stuck" is "unstuck",  (removing both the class which sets fixed positioning and the shim).

This PR adds mocks for the window scroll position, offset of the sticky element and the window dimensions and updates stickAtTopWhenScrolling to use these.

It adds new tests to ensure that the three conditions of checkScroll are met, to "stick" or "unstick (release)" a sticky element.

Tests:

screen shot 2016-09-20 at 17 57 03

PR #335, opened 2016-10-07
v4.19.0

gemmaleigh:
Version 4.19.0 adds the following:

[PR #329](https://github.com/alphagov/govuk_frontend_toolkit/pull/329)

Amend GOVUK.StickAtTopWhenScrolling to check on resize if an element
should be “unstuck”.

For smaller screens (<768px) ensure that the
GOVUK.StickAtTopWhenScrolling JS "unsticks" the element which was
previously "stuck" (by removing both the class which sets fixed
positioning and the shim).


gemmaleigh:
Closing in favour of #336.

PR #339, opened 2016-10-19
Add semicolons to the start of each js file

gemmaleigh:
Add a leading semicolon in front of immediately-invoked function
expressions, so that each file can be safely concatenated.


nickcolley:
Thanks @gemmaleigh 👍 

We should also add this to https://github.com/alphagov/styleguides/blob/master/js.md#modules

PR #342, opened 2016-10-20
Remove semicolons from JS documentation

gemmaleigh:
These no longer exist in the codebase, this PR updates these examples to match.

cc. @nickcolley, @robinwhittleton.

This should probably be merged before #341.

### DIFF
--- a/docs/javascript.md
+++ b/docs/javascript.md
@@ -23,7 +23,7 @@ Modules can be found and started by including `govuk/modules.js` and running:
 
 ```javascript
 $(document).ready(function(){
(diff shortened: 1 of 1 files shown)


docs/javascript.md
```diff
@@ -56,7 +56,7 @@ The simplest module looks like:
       // module code
     }
   };
-})(window.GOVUK.Modules);
+})(window.GOVUK.Modules)
```
nickcolley: We need to add the semi at the start of the IIFE I think?

PR #343, opened 2016-10-24
Amend GOVUK.StickAtTopWhenScrolling 

gemmaleigh:
Add a check for a .js-sticky-resize class

If it exists on the sticky element, then find the width of the parent
element (as we’re assuming by having set this class that it’s in a
grid-column that will change it’s width when resized).

Get the width of the parent element and set the width of the shim and
the width of the sticky element to match when the window is resized.

In the tests, create a wrapper element with a width defined (300px) to
check that the width is also set for the sticky element and shim.

This is needed for service-manual-frontend and can be [previewed over here](https://github.com/alphagov/service-manual-frontend/pull/61).

### DIFF
--- a/javascripts/govuk/stick-at-top-when-scrolling.js
+++ b/javascripts/govuk/stick-at-top-when-scrolling.js
@@ -91,6 +91,15 @@
         sticky.$els.each(function (i, el) {
           var $el = $(el)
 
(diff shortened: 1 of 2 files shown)


robinwhittleton:
Yep, I’m happy with this.

PR #347, opened 2016-10-28
Fix role click shim

nickcolley:
When trying to implement the shim I found the page will scroll when pressing space, then you navigate to the page.

To fix this we need to instead listen for `keydown` and not `keyup` so we can prevent the unwanted scroll behaviour.

I've also added a commit to simplify the shim since there's not a clear need for this to be flexible.

### DIFF
--- a/javascripts/govuk/shim-links-with-button-role.js
+++ b/javascripts/govuk/shim-links-with-button-role.js
@@ -1,13 +1,11 @@
 // javascript 'shim' to trigger the click event of element(s)
 // when the space key is pressed.
 //
-// usage instructions:
-// GOVUK.shimLinksWithButtonRole.init();
+// Created since some Assistive Technologies (for example some Screenreaders)
+// Will tell a user to press space on a 'button', so this functionality needs to be shimmed
 //
-// If you want to customise the shim you can pass in a custom configuration
-// object with your own selector for the target elements and addional keyup
-// codes if there becomes a need to do so. For example:
-// GOVUK.shimLinksWithButtonRole.init({ selector: '[role="button"]' });
+// Usage instructions:
+// GOVUK.shimLinksWithButtonRole.init();
 ;(function (global) {
   'use strict'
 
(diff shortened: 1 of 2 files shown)


javascripts/govuk/shim-links-with-button-role.js
```diff
@@ -16,40 +14,16 @@
...
-      // if we have found elements then:
-      if ($(this.config.selector).length > 0) {
-        // listen to 'document' for keyup event on the elements and fire the triggerClickOnTarget
-        $(document).on('keyup', this.config.selector, this.triggerClickOnTarget.bind(this))
-      }
+    init: function init () {
+      // listen to 'document' for keydown event on the any elements that should be buttons.
+      $(document).on('keydown', '[role="button"]', function (event) {
+        // if the keyCode (which) is 32 it's a space, let's simulate a click.
+        if (event.which === 32) {
```
tvararu: Can be rewritten as:

```js
var userPressedSpace = event.which === 32
if (userPressedSpace) {
```

Wouldn't need the comment then.

javascripts/govuk/shim-links-with-button-role.js
```diff
@@ -1,13 +1,11 @@
 // javascript 'shim' to trigger the click event of element(s)
 // when the space key is pressed.
 //
-// usage instructions:
-// GOVUK.shimLinksWithButtonRole.init();
+// Created since some Assistive Technologies (for example some Screenreaders)
```
robinwhittleton: Is this true? As I understand it this isn’t about assistive tech but about normalising the behaviour of real buttons and links-styled-to-look-like-buttons but standard browsers.

nickcolley: I thought this was the reason we did this, since some AT are telling users to press space which at the moment does not work?

robinwhittleton: @cfq arbitration needed 🙂

gemmaleigh: I'm with @nickcolley on this one. 

From the original thread: 
https://github.com/alphagov/govuk_elements/pull/272
--

@aduggin said:

This is quite a complex issue - requires a long blog post to explain it fully. 

**The short version**
For links that have role="button" we need to add an event handler so that it can be activated with the space bar (https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_button_role)

**The medium version**
We have links and buttons that look the same so there is a consistent call to action for sighted users.
Unfortunately this creates inconsistency for screen reader users and speech recognition users.
To progress through a transaction they have to sometimes interact with a link and sometimes with a button.

Adding role="button" to links provides consistency as it makes screen readers and speech recognition software treat it as a button. This only works in some screen reader browser combinations and the latest version of Dragon Naturally speaking (13) that support role="button".

Buttons can be activated with the space bar, links can't. This behaviour needs to be added with javascript to a link with role="button" as it is not automatically added by adding role="button"

So, adding role="button" and an event handler for the space bar will only provide the required consistency to some assistive technology users.

In an ideal world links look like links and buttons look like buttons

**The long version**
In a blog post that I need to find time to write - but includes information about how screen reader users and dragon users interact with links and buttons and the scenarios of when people will be confused by our links and buttons looking the same.

nickcolley: I've added a link to Alistair's comment for context

cfq: Thanks @gemmaleigh, that's the whole answer. We started with AT users but our investigation has shown us that browser vendors haven't implemented the functionality on purpose and this was the recommended solution.

cfq: Here's a video of our test on buttons vs links: https://www.youtube.com/watch?v=NO7u2tEbFQY

robinwhittleton:
I thought I remembered something about `keydown` being a bit weird in old IE, but I think I was actually thinking about `keypress`. I guess this has been tested in IE8+? If so I’ll merge.

nickcolley:
Okay after testing this again I realise keyup is the right behaviour for space.

So this PR needs to be changed to handle `keydown` to prevent scrolling of the page, and handle `keyup` to click the button...

nickcolley:
I can't find a place where that behaviour is defined in any spec, so I'm presuming it's down the the operating system.

I've tested this PR in IE8,IE9,IE10 and it works.

So this should be good to go as long as we can stomach the spacebar submitting on keydown and not keyup.

javascripts/govuk/shim-links-with-button-role.js
fofr: Rather than the comment I’d have written something like: `var SPACEBAR = 32;` and `if (event.which === SPACEBAR) {` etc.

PR #348, opened 2016-11-07
Make font variables lowercase

selfthinker:
Our sass linting only allows hyphenated lowercase names. As all of our font variable names have uppercase letters, the `NameFormat` option needs to be disabled when they are used.
This makes all those variables lowercase while also providing a fallback uppercase name to keep it backwards compatible so that we can enable that linting option again in the future.

### DIFF
--- a/stylesheets/_font_stack.scss
+++ b/stylesheets/_font_stack.scss
@@ -1,19 +1,25 @@
 //  GOV.UK font stacks, referred to in typography.scss
 
-// Allow uppercase letters in font stack variable names
-// scss-lint:disable NameFormat
-
 // New Transport Light
-$NTA-Light: "nta", Arial, sans-serif;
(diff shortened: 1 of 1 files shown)


stylesheets/_font_stack.scss
```diff
@@ -4,16 +4,23 @@
 // scss-lint:disable NameFormat
```
nickcolley: If we do this can we move this scss-lint below where we're using the caps?

selfthinker: Yes, makes sense.
I'm not sure but if I understand the docs correctly, that disables it for the entire file anyway, independent of the line being at the top or at the bottom. But even if that's the case, it still makes sense to have it closer to what it refers to.

fofr:
We should version this as a breaking change.

robinwhittleton:
If @selfthinker hadn’t left a mapping at the bottom of the file back to the old spellings I’d agree with you :)

PR #350, opened 2016-11-22
Better selection button docs

robinwhittleton:
The documentation for selection buttons was in a pretty bad state. This tidies the code examples and attempts to do a much better job at explaining how delegated eventing for selection buttons works.


gemmaleigh:
Thanks @robinwhittleton & @abbott567!

PR #353, opened 2016-11-24
Change colour used in phase tags to govuk-blue

edwardhorsford:
## Summary
This pr changes the background colour used in the phase tags to $govuk-blue rather than $alpha-colour or $beta-colour which have insufficient contrast.

The current pink has a contrast of 4.46:1 (nearly ok), but the orange is 2.78:1.

Have discussed with accessibility team and @timpaul but will want to discuss with the rest of the design team.

Ideally I'd use a semantic sass name, but I believe $mainstream-brand may change per org, which is not what we want here.

Prints a sass warning to let users know passing a phase tag is depreciated.

Before:
screen shot 2016-11-24 at 18 06 58
screen shot 2016-11-24 at 18 07 03

After:
screen shot 2016-11-24 at 18 07 24
screen shot 2016-11-24 at 18 07 37

## Why not adjust to darker colours?

The pink can be adjusted (will do regardless in a separate pr), but there's no way of getting the orange to meet contrast checks without being brown or red. 

Separately - there's no real reason for them to have different colours - users don't know what alpha or beta mean, let alone the colours. Changing them to blue makes the pages more consistent.

## Why not remove the banners entirely?

Is a possible solution, but will need much more discussion. This is an easier change to do - whilst any other discussion about keeping or removing the banners happens.

## Testing

These changes haven't been tested - am unsure how to test frontend toolkit changes.

## Still to do / oustanding:
* need to update elements / other bits of GOV.UK that use the phase banner.

### DIFF
--- a/docs/mixins.md
+++ b/docs/mixins.md
@@ -534,16 +534,12 @@ A mixin to create a GOV.UK Phase banner, with alpha/beta tag inside.
 
 #### Description
 
-`@mixin phase-banner($state)`
-
-`$state` is either `alpha` or `beta`.
-
-`$state` sets the background colour of the phase tag to the appropriate alpha or beta colour.
+`@mixin phase-banner()`
 
 ##### Phase banner - Alpha
 
     .phase-banner  {
-      @include phase-banner(alpha);
+      @include phase-banner();
     }
 
     <div class="phase-banner">
@@ -556,7 +552,7 @@ A mixin to create a GOV.UK Phase banner, with alpha/beta tag inside.
 ##### Phase banner - Beta
 
     .phase-banner  {
-      @include phase-banner(beta);
+      @include phase-banner();
     }
 
     <div class="phase-banner">
@@ -572,16 +568,12 @@ A mixin to create an alpha/beta tag.
 
 #### Description
 
-`@mixin phase-tag($state)`
-
-`$state` is either `alpha` or `beta`.
-
-`$state` sets the background colour of the phase tag to the appropriate alpha or beta colour.
+`@mixin phase-tag()`
 
 ##### Phase tag - Alpha
 
     .alpha-tag{
-      @include phase-tag(alpha);
+      @include phase-tag();
     }
     <h2>
       Apply using the new service <span class="alpha-tag">ALPHA</span>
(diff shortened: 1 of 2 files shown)


edwardhorsford:
Fixes govuk_elements/#217

nickcolley:
A first thought is since we're using a similar colour to our links, this could make the new label look like it's a link and therefore clickable? Just an assumption.

robinwhittleton:
As it stands this will be a breaking change. Calling `@include phase-banner($alpha)` with the new code breaks Sass compilation as the mixin isn’t expecting an argument. Probably the best solution is to change the definition to `@mixin phase-banner($phase: '')` to imply that you’re not expecting a value, that way regardless of how it’s called it will work.

fofr:
Will anything else be using `$alpha-colour` and `$beta-colour`.
If not, they can be removed. If yes – what are those things, will they also need updating?

edwardhorsford:
@robinwhittleton thanks, will adjust the PR.

@fofr I think they're used in other places too - right now I'm not suggesting removing the semantic colours, just changing this pattern to no longer use them.

@nickcolley I don't think it's an issue. @timpaul has had these in some research lately - perhaps he can say more.

fofr:
Why blue? Why not make both banners pink or a different non-blue colour? There may not be a need to have different colours in the banner, but I think there’s a need for it to flag users, by appearing differently to the rest of GOV.UK.

Here's the change in context:
screen shot 2016-11-25 at 15 09 51
screen shot 2016-11-25 at 15 10 14

edwardhorsford:
@fofr I don't think any other colour will be more 'noticeable'. Already we see that the banners aren't noticed my users in research. They're more for service teams in government to know the status of the project.

@timpaul any thoughts?

timpaul:
The new style featured in some recent research into GaaP product pages and didn't cause any issues, but we weren't explicitly testing them. 

I agree with Ed that most users already have very little awareness or understanding of what these tags mean - and that's fine. It's the messaging next to the tag and on service start pages that's important. In terms of improving accessibility and reducing visual noise I think the blue works.

edwardhorsford:
Ok, have spoken to @markhurrell who is happy with this change also. I'm going to remove the [DO NOT MERGE] label.

After this is merged, I think we'll need to:
* Update service manual with new examples
* Bump alphagov/govuk-elements to include latest kit
* I think they're used on a few gov.uk content pages - hope these just need a bump of the kit to get the new styles.

robinwhittleton:
This looks good to me.

PR #354, opened 2016-11-29
v5.0.2

gemmaleigh:
Change the phase banner and phase tag mixins to no longer require a state, as the phase tag background colour is now govuk-blue ([PR #353](https://github.com/alphagov/govuk_frontend_toolkit/pull/353)).


robinwhittleton:
Bump patch level as it’s a bugfix for accessibility contrast issues.

PR #356, opened 2016-12-05
setInitialState supports both string and jQuery

### LINKED ISSUE #355: GOVUK SelectionButtons setInitialState requires jQuery object instead of list of elements
Should the SelectionButtons setInitialState method be less restrictive on the attributes it accepts?

https://github.com/alphagov/govuk_frontend_toolkit/blob/master/javascripts/govuk/selection-buttons.js#L33

I was following https://github.com/alphagov/govuk_frontend_toolkit/blob/master/docs/javascript.md#selection-buttons and the usage of setInitialState when I discovered that it needed a jQuery Object instead of a selector string to loop through. 

I'm happy to either update the guidance and/or propose a change to help support both jQuery and selector string but wondered what others thought about it.

aliuk2012:
fixes #355

setInitialState only accepted jQuery object so if you
passed a string of selectors it would break. My changes
tests if the parameter value is an instance of
jQuery and if it is not then it attempts to convert it
to one by calling $();

### DIFF
--- a/javascripts/govuk/selection-buttons.js
+++ b/javascripts/govuk/selection-buttons.js
@@ -16,7 +16,7 @@
     }
     if (typeof elmsOrSelector === 'string') {
       this.selector = elmsOrSelector
-      this.setInitialState($(this.selector))
+      this.setInitialState(this.selector)
     } else if (elmsOrSelector !== undefined) {
       this.$elms = elmsOrSelector
       this.setInitialState(this.$elms)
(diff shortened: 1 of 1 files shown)


javascripts/govuk/selection-buttons.js
```diff
@@ -30,7 +30,9 @@
       this.addDocumentLevelEvents()
     }
   }
-  SelectionButtons.prototype.setInitialState = function ($elms) {
+  SelectionButtons.prototype.setInitialState = function (elmsOrSelector) {
+    var $elms = elmsOrSelector instanceof $ ? elmsOrSelector : $(elmsOrSelector)
```
aliuk2012: @colinrotherham agreed, I've updated this pr.

robinwhittleton:
Can we add a test or two around this new functionality?

aliuk2012:
Sure I'll add some tests to the new functionality and for the existing setInitialState

nickcolley:
[Release v5.2.0](https://github.com/alphagov/govuk_frontend_toolkit/releases/tag/v5.2.0) deprecates this JavaScript so I think this can be closed.

Thanks @aliuk2012 for contributing 👍  💯

PR #357, opened 2016-12-06
Allow tablet, desktop breakpoints to be overridden

36degrees:
We are using GOV.UK Frontend Toolkit in the tech docs work we are doing (although it’s currently just vendor’ed in) and need to override the breakpoints used by the `media` mixin so that the behaviour makes sense for our application.

This makes the tablet and desktop breakpoints variables with default values, which can be overriden by e.g. specifying different values in the application’s stylesheet before frontend toolkit is included.

### DIFF
--- a/stylesheets/_conditionals.scss
+++ b/stylesheets/_conditionals.scss
@@ -25,6 +25,9 @@
 $is-ie: false !default;
 $mobile-ie6: true !default;
 
+$tablet-breakpoint: 641px !default;
+$desktop-breakpoint: 769px !default;
+
 @mixin media($size: false, $max-width: false, $min-width: false, $ignore-for-ie: false) {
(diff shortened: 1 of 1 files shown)


stylesheets/_conditionals.scss
```diff
@@ -34,15 +37,15 @@ $mobile-ie6: true !default;
...
         @content;
       }
     } @else if $size == tablet {
-      @media (min-width: 641px){
+      @media (min-width: $tablet-breakpoint){
         @content;
       }
     } @else if $size == mobile {
-      @media (max-width: 640px){
+      @media (max-width: $tablet-breakpoint - 1px){
```
fofr: Should `$mobile-breakpoint` be defined as `$tablet-breakpoint - 1px`?

36degrees:
Thanks, @robinwhittleton!

PR #359, opened 2016-12-12
[DO NOT MERGE] Add tracker to track clicks on breadcrumb links

carvil:
For review only, this still needs validation from performance analysts so we know it's doing the expected behaviour.

------

This new tracker will track clicks on breadcrumb links and collect the
following information in a custom event:

- The event category: `breadcrumbClicked`;
- The action: the position of the link in the breadcrumbs (e.g. 2);
- The label: the href to which the link points to (e.g. `/browse/benefits/entitlement`);
- The link text as a custom dimension (e.g. `Benefits entitlement`).

The custom dimension is set prior to calling the `trackEvent` function, so it's sent in that event.

**NOTE**: the custom dimension value is still undetermined, the existing
one is a placeholder and needs to be replaced before this goes live.

Trello:
https://trello.com/c/buYUPz1l/247-enable-tracking-of-breadcrumb-clicks-across-applications-rendering-education-content

### DIFF
--- a/javascripts/govuk/analytics/breadcrumb-link-tracker.js
+++ b/javascripts/govuk/analytics/breadcrumb-link-tracker.js
@@ -1,0 +1,50 @@
+;(function (global) {
+  'use strict'
+
+  var $ = global.jQuery
+  var GOVUK = global.GOVUK || {}
+
+  GOVUK.analyticsPlugins = GOVUK.analyticsPlugins || {}
+  GOVUK.analyticsPlugins.breadcrumbLinkTracker = function (options) {
+    options = options || {}
+    var breadcrumbLinkSelector = options.selector
+
+    if (breadcrumbLinkSelector) {
+      $('body').on('click', breadcrumbLinkSelector, trackBreadcrumbClick)
+    }
+
+    function trackBreadcrumbClick (evt) {
+      var linkPosition = $(breadcrumbLinkSelector).index(this)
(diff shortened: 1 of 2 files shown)


fofr:
There's some prior art elsewhere which I think we should re-use:
https://github.com/alphagov/frontend/blob/master/app/assets/javascripts/modules/track-click.js

carvil:
@fofr thanks for your comments. The reason for adding it here is because there are already a few trackers in the repo that do similar things. But I can definitely move it to static instead, if that's the preferred way.

This tracker is slightly different than `track-click`, though. We need to set a custom dimension first before we track the custom event. I am tempted to write a similar module (e.g. `track-breadcrumb-click` with that extra logic rather than applying conditionals to it. What do you think of that?

Alternatively I can copy the `track-click` module to static and change it to allow custom dimensions too, but that will not apply to the majority of the cases.

fofr:
Is it possible to port `track-click` and create a `track-breadcrumb-click` that sets the custom dimension and then calls `track-click` somehow?

This is an example of an admin module doing something similar:
https://github.com/alphagov/publisher/blob/master/app/assets/javascripts/modules/ajax_save_with_parts.js#L5-L9

carvil:
Thanks @fofr . Closing this PR in favour of:

- https://github.com/alphagov/static/pull/863
- https://github.com/alphagov/static/pull/864

PR #362, opened 2016-12-29
Use Travis to push and trigger downstream jobs

alexmuller:
The Jenkins that this project was previously using is now just for GOV.UK Publishing projects, so this repo has to start using Travis for builds.

This change adds a deploy key to the GitHub repo and encrypts the private key for Travis to use to push commits.

It also adds my encrypted Travis API key to the environment to allow us to trigger downstream jobs for the gem and npm packages.


fofr:
Will merging this allow us to publish new versions of the toolkit again?

PR #365, opened 2017-01-10
Allow custom options when tracking events

alecgibson:
Adding the ability to track custom dimensions so that we can define them in the `options` object, and perform calls like `GOVUK.analytics.trackEvent('category', 'action', {dimension29: 'dimensionValue'});`

This is required by markups suggested here: https://github.com/alphagov/static/pull/875#pullrequestreview-15881444

### DIFF
--- a/javascripts/govuk/analytics/google-analytics-universal-tracker.js
+++ b/javascripts/govuk/analytics/google-analytics-universal-tracker.js
@@ -102,7 +102,26 @@
       evt.transport = options.transport
     }
 
+    var customDimensions = getCustomDimensions(options)
+    if (customDimensions) {
+      $.extend(evt, customDimensions)
+    }
+
     sendToGa('send', evt)
+
+    function getCustomDimensions(options) {
+      var customDimensionMatcher = /^dimension[0-9]+$/;
+      var customDimensions = null
+
+      $.each(options, function(key, value) {
+        if (customDimensionMatcher.test(key)) {
+          customDimensions = customDimensions || {}
+          customDimensions[key] = value
+        }
+      })
+
+      return customDimensions
+    }
   }
 
   /*
--- a/javascripts/govuk/analytics/google-analytics-universal-tracker.js
+++ b/javascripts/govuk/analytics/google-analytics-universal-tracker.js
@@ -72,11 +72,7 @@
     // Label is optional
     if (typeof options.label === 'string') {
       evt.eventLabel = options.label
-    }
-
-    // Page is optional
-    if (typeof options.page === 'string') {
-      evt.page = options.page
+      delete options.label
     }
 
     // Value is optional, but when used must be an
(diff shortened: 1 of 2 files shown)


fofr:
I'd prefer an approach similar to https://github.com/alphagov/govuk_frontend_toolkit/pull/332
This would allow an options object to be passed onto GA itself.

alecgibson:
@fofr Do you mean just passing the `options` object straight through? Fine by me - is there a reason this wasn't done before? I'd sort of figured it was a sanitisation thing? Or do we just trust ourselves to not pass anything silly in?

fofr:
@alecgibson Yep, straight through.

There was a concept of having analytics providers, so google-analytics-universal-tracker is one such provider. There used to be two – the other was google-analytics-classic-tracker. This pattern allowed us to send the same data to both even though they had different APIs. That's not the case any more.

alecgibson:
@fofr Great will do. There's a couple of annoying parameter mismatches (eg `label` => `eventLabel`) which changing would presumably break a bunch of other stuff? I'm guess I should just leave these for now?

javascripts/govuk/analytics/google-analytics-universal-tracker.js
```diff
@@ -85,21 +81,19 @@
...
-      evt.nonInteraction = 1
+      options.nonInteraction = 1
     }
 
-    // Set the transport method for the event
-    // Typically used for enabling `navigator.sendBeacon` when the page might be unloading
-    // https://developers.google.com/analytics/devguides/collection/analyticsjs/field-reference#transport
-    if (options.transport) {
-      evt.transport = options.transport
+    if (typeof options === 'object') {
```
fofr: Is this check necessary given line 64?

spec/unit/analytics/google-analytics-universal-tracker.spec.js
```diff
@@ -100,6 +100,13 @@ describe('GOVUK.GoogleAnalyticsUniversalTracker', function () {
       )
     })
 
+    it('tracks custom dimensions', function() {
+      universal.trackEvent('category', 'action', {dimension29: 'Home'})
+      expect(window.ga.calls.mostRecent().args).toEqual(
+        ['send', {hitType: 'event', eventCategory: 'category', eventAction: 'action', dimension29: 'Home'}]
```
fofr: Have you confirmed that this object structure correctly passes the dimension to that event hit?

alecgibson: I've now tested this on dev with a forked Frontend Toolkit gem, and a WIP Static branch that sends custom dimensions. All looks ok.

fofr:
A side effect of this change is that through code inspection the `page` and `transport` event features are no longer visible, instead only discoverable via the GA api docs.

alecgibson:
@fofr yes that's the trade-off, but if we want to allow _anything_ from the GA API, I don't think it's unreasonable to expect people to follow the docs.

alecgibson:
@fofr I've now tested this end-to-end with https://github.com/alphagov/static/pull/875

I've verified that the event turns up in the Integration/Staging account of Google Analytics. Good to merge?

PR #370, opened 2017-01-13
For discussion: Make alpha, beta and discovery colours $govuk-blue

gemmaleigh:
The phase tag is now $govuk-blue (as previously the pink and orange
colours used failed contrast requirements).

Do we want to update the Sass variables for $alpha-colour, $beta-colour and $discovery-colour to also use $govuk-blue?

Opened as this related issue was raised with the service manual: https://github.com/gds-attic/government-service-design-manual/issues/671

### DIFF
--- a/stylesheets/colours/_palette.scss
+++ b/stylesheets/colours/_palette.scss
@@ -68,9 +68,9 @@ $panel-colour: $grey-3;           // Related links panel, page footer etc.
 $canvas-colour: $grey-4;          // Page background
 $highlight-colour: $grey-4;       // Table stripes etc.
 $page-colour: $white;             // The page
-$discovery-colour: $fuschia;      // Discovery badges and banners
-$alpha-colour: $pink;             // Alpha badges and banners
-$beta-colour: $orange;            // Beta badges and banners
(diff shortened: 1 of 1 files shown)


accessiblewebuk:
short answer - yes

longer answer - yes, but maybe keep the status colours category, with just the $error-colour - although that red is in the main palette as well, it probably helps to make it clear that error messages and status bars should be that colour.

accessiblewebuk:
Is this also about changing page http://govuk-elements.herokuapp.com/colour - if so then we should remove the status colours from that page, apart from $error-message (even though that red is in the main palette, it helps people know that it should be used for error messages and status bars.

gemmaleigh:
It's partly about changing the Colour page for elements. 

That page can be updated without affecting the frontend toolkit, so I've made those changes:
https://github.com/alphagov/govuk_elements/pull/385

I'm also concerned about there being references to old colours that aren't being used any more. It would be good if there was a way to deprecate these, if we can agree that they should only be used for printed materials.

gemmaleigh:
@tvararu I was thinking of something like this! https://github.com/salesforce-ux/sass-deprecate

PR #374, opened 2017-01-18
Add colour contrast test

### LINKED ISSUE #316: Department colours are not always accessible
Several department colours do not have sufficient contrast, and are being used for text.

Example: [Defra org page](https://www.gov.uk/government/organisations/department-for-environment-food-rural-affairs)

Departments with issues:
- DEFRA
- HMRC
- Any child orgs that inherit Defra or HMRC colours.

Note I think there's several issues:
- All departments should have a `contrast-safe` or `text-contrast-safe` colour.
- All departments should use these colours for text.
- We currently have `web-safe` colour variants to mean contrast safe, but this does not communicate the purpose of the variant, and has led to some being deleted - such as #249. Web-safe means something different.

Update:
The issue with `web-safe` is that it has a [specific meaning](https://en.wikipedia.org/wiki/Web_colors). It's also been used to refer to a hex version of another named colour. Such as when an org has a Pantone colour as their brand colour, and a 'web version' for internet use.

For our use, all the hex colours can be used on the web, but only some will have suitable contrast for text.

We must meet WCAG 2.0 AA, which requires a contrast against the background of 4.5:1 or greater.

fofr:
Test that all organisation colours have a contrast ratio of 4.5 or greater against white. Error when contrast too low.

* Protects against poor colour choices in future
* Use sass-color-helpers to provide contrast function
  https://github.com/voxpelli/sass-color-helpers
* Only include colour helpers when in dev, don't package with gem
* Store sass test in same directory as generated sass test file
* Confirms that all current colours have good contrast ratio
* Confirms that old DEFRA colour was poor contrast ratio

Fixes #316

Example failing test:
https://travis-ci.org/alphagov/govuk_frontend_toolkit/builds/193155272

```
Error: Contrast ratio for department-for-environment-food-rural-affairs too low.
#00a33b on #fff has a contrast of: 3.3. Must be higher than 4.5 for WCAG AA support
        on line 10 of spec/stylesheets/_colour_contrast_spec.scss
        from line 18 of ./spec/stylesheets/test.scss
```

cc @robinwhittleton @nickcolley @edwardhorsford 

Thanks to @voxpelli for making this easy!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,5 @@
 node_modules/
 .sass-cache
 npm-debug.log
-spec/stylesheets
+spec/stylesheets/test*
 _SpecRunner.html
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -16,6 +16,8 @@ module.exports = function (grunt) {
     }
   )
 
+  allSassFiles.push("@import '_colour_contrast_spec.scss';")
+
   grunt.file.write(
     './spec/stylesheets/test.scss',
     allSassFiles.join('\n')
@@ -46,7 +48,9 @@ module.exports = function (grunt) {
         },
         options: {
           loadPath: [
-            './stylesheets'
+            './node_modules',
+            './stylesheets',
+            './spec/stylesheets'
           ],
           style: 'nested'
         }
--- a/package.json
+++ b/package.json
@@ -9,7 +9,8 @@
     "grunt-contrib-jasmine": "^1.0.0",
     "grunt-contrib-sass": "0.7.4",
     "jquery": "~1.12.4",
-    "standard": "^8.2.0"
+    "standard": "^8.2.0",
+    "sass-color-helpers": "^2.0.3"
   },
   "scripts": {
     "test": "grunt test && npm run lint --silent",
--- a/spec/stylesheets/_colour_contrast_spec.scss
+++ b/spec/stylesheets/_colour_contrast_spec.scss
@@ -1,0 +1,12 @@
+@import "colours/organisation";
+@import "sass-color-helpers/stylesheets/color-helpers/math";
+@import "sass-color-helpers/stylesheets/color-helpers/contrast";
+
+@each $organisation in $all-organisation-brand-colours {
+  $minimum-contrast: 4.5;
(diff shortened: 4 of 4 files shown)


edwardhorsford:
This is really cool! Thanks @fofr!

markhurrell:
hey @fofr this is a great idea. can we also run the contrast test against black/text colour to make sure the colour is suitable for links

edwardhorsford:
Talking with @accessiblewebuk he points out that if you're going to use colour alone to indicate a link, it [should have 3:1 against text colour](https://www.w3.org/TR/2016/NOTE-WCAG20-TECHS-20161007/G183).

accessiblewebuk:
If you indicate link text by colour alone then it needs to contrast with the other text by at least 3:1 - the real difficulty is that both texts need to contrast with the background by at least 4.5:1

see https://www.w3.org/TR/2016/NOTE-WCAG20-TECHS-20161007/G183

accessiblewebuk:
You can get around the "indicating by colour alone" by making the link text bold compared to surrounding text and/or underlining it.

markhurrell:
@accessiblewebuk true but text styles like bold and underline can be contextually inappropriate in things like long, nested navigation pages etc (which GOV.UK has quite a lot of)

as we're implementing something that gives hard yes/no for accessibility it would be good to make sure it is actually accessible for those use-cases

fofr:
@markhurrell Good call on the other colours we use. I think that'd be a separate PR.

fofr:
@edwardhorsford @markhurrell @accessiblewebuk Want to raise an issue discussing how to do link-colour contrast checking? It's an interesting point – I'm sure some of these organisation colours, used for links, don't have a 3:1 ratio with black.

edwardhorsford:
@markhurrell You could do it as a `warn` rather than `error` so that it doesn't give a hard no.

accessiblewebuk:
@markhurrell - do you have some examples of different use cases? Navigation normally has other contextual information like it is in a list, or in blocks. The 3.1 is mainly to do with text links embedded in body copy. I don't know the maths but there are probably only a very narrow range of combinations of three colours that work (text, link text and background).

nickcolley:
Looks good to me 👍

PR #376, opened 2017-01-30
Fix a weird stick at top when scrolling bug

36degrees:
For _reasons unknown_, getting the width of the 'stuck' element by calling `$el.width()` _after_ adding the shim to the document causes Chrome 56 to scroll down by the height of the shim, effectively ‘jumping’ the user down the document. This happens even if you just e.g. call `console.log($el.width)` without using the returned value.

This behaviour does not occur in other browsers, nor did it occur in Chrome 55.

Somehow, this fixes it.

¯\\_(ツ)_/¯

## Before
before

## After
after

### DIFF
--- a/javascripts/govuk/stick-at-top-when-scrolling.js
+++ b/javascripts/govuk/stick-at-top-when-scrolling.js
@@ -110,8 +110,9 @@
       if (!$el.hasClass('content-fixed')) {
         $el.data('scrolled-from', sticky.getElementOffset($el).top)
         var height = Math.max($el.height(), 1)
-        $el.before('<div class="shim" style="width: ' + $el.width() + 'px; height: ' + height + 'px">&nbsp;</div>')
-        $el.css('width', $el.width() + 'px').addClass('content-fixed')
+        var width = $el.width()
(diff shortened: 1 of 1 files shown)


robinwhittleton:
Calling `.width()` in jQuery against an element does trigger a reflow, but I’m not sure what that’s triggering the bug. Either way, even without the bug fix this is a microptimisation so let’s get it in.

Are you going to file a Chrome bug?

36degrees:
Struggling a little to define what the buggy behaviour actually _is_. I'm trying to recreate the bug in a codepen but without much luck so far… If I can get a better test case then I will file a bug.

nickcolley:
Related: https://github.com/alphagov/static/issues/886

36degrees:
@nickcolley I looked at using `position: sticky` with a polyfill, but whilst the native browser behaviour does work perfectly where implemented, none of the existing polyfills worked well for this use case. The main issue is that we only want the sticky behaviour on a `> tablet` viewport, and none of the polyfills I could find had any consideration for media queries.

36degrees:
[Bug filed against Chromium](https://bugs.chromium.org/p/chromium/issues/detail?id=687118)

nickcolley:
@36degrees my suggestion is we do not polyfill this since we'll run into issues like this in the future also.

PR #379, opened 2017-02-13
Breadcrumb anchor link styling

alecgibson:
This style any breadcrumbs linking to anchors on the current page. This is used in conjunction with https://github.com/alphagov/govuk_navigation_helpers/pull/30 to show users which page they are currently on, when browsing the new taxonomy.

Before:
govuk_frontend_toolkit-current-breadcrumb-before

After:
govuk_frontend_toolkit-current-breadcrumb-after

Trello: https://trello.com/c/MVM5bueS/386-add-current-page-to-taxonomy-breadcrumbs

### DIFF
--- a/stylesheets/design-patterns/_breadcrumbs.scss
+++ b/stylesheets/design-patterns/_breadcrumbs.scss
@@ -48,6 +48,11 @@
 
   a {
     color: $text-colour;
+
+    &[href^="#"] {
+      color: $secondary-text-colour;
+      text-decoration: none;
+    }
   }
 
 }


alecgibson:
As pointed out by @tijmenb if people think it's more appropriate, this change could live in `static` rather than `govuk_frontend_toolkit`.

nickcolley:
Can you add details of the research that led to this change?

alecgibson:
Moving this change to static: https://github.com/alphagov/static/pull/901

PR #383, opened 2017-02-20
Allow showing and hiding of multiple targets

pcraig3:
### Allow showing/hiding of more than one target

The WC3 spec says that aria-controls should accept an ID
reference list, but our show/hide logic had been written in a way
that assumed only one id.

Extending the behaviour so that we can pass in a space-separated
string of ids to the "data-target" attribute and show/hide
multiple elements at once.

Source: https://www.w3.org/TR/wai-aria/states_and_properties#aria-controls

I can add something about this to the README if we're happy with it.

***

multi

### DIFF
--- a/javascripts/govuk/show-hide-content.js
+++ b/javascripts/govuk/show-hide-content.js
@@ -24,30 +24,48 @@
     function initToggledContent () {
       var $control = $(this)
       var $content = getToggledContent($control)
+      var contentIds = ''
 
       // Set aria-controls and defaults
       if ($content.length) {
-        $control.attr('aria-controls', $content.attr('id'))
+        $content.each(function () {
+          $(this).attr('aria-hidden', 'true')
+          contentIds += ' ' + $(this).attr('id')
+        })
+
+        $control.attr('aria-controls', contentIds.trim())
         $control.attr('aria-expanded', 'false')
-        $content.attr('aria-hidden', 'true')
       }
     }
 
     // Return toggled content for control
     function getToggledContent ($control) {
-      var id = $control.attr('aria-controls')
+      var targetIds = $control.attr('aria-controls')
 
       // ARIA attributes aren't set before init
-      if (!id) {
-        id = $control.closest('label').data('target')
+      if (!targetIds) {
+        targetIds = $control.closest('label').data('target')
+      }
+
+      // turn a space-separated list of ids into a comma-separated css id selector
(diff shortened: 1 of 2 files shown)


edwardhorsford:
Interesting! To play devil's advocate, I'm going to ask if this is behaviour we want to support / encourage - @timpaul?

We'd normally suggest to teams that they not be revealing lots / multiple things. Conditional questions should clearly be a 'child' of the parent question. In this case I'd have the left grey border span both elements.

dsingleton:
Hi @pcraig3, thanks for the contribution.

As we discussed on Slack theres an issue here with our Travis setup that means [builds fail](https://travis-ci.org/alphagov/govuk_frontend_toolkit/builds/203551725) for anyone submitting a PR via a fork.

Our [travis config](https://github.com/alphagov/govuk_frontend_toolkit/blob/master/.travis.yml#L5-L10) includes some setup/configuration for deployment ([more detail here](https://github.com/alphagov/govuk_frontend_toolkit/tree/master/.travis)) that depend on secret values, which are limited to users with write permission to this repo.

I think we can fix this by moving the deployment configuration steps from `before_install` to `before_deploy`, as the [later is only run when a deployment is triggered](https://docs.travis-ci.com/user/deployment/npm/#Running-commands-before-and-after-deploy), which shouldn't ever happen for a fork.

I'm sorting out a few other Travis issues at the moment, and i'm try to fix this next. 

Let me know if this becomes a blocker for you, and apologies for the problems!

dsingleton:
Hi @pcraig3, 

Once https://github.com/alphagov/govuk_frontend_toolkit/pull/385 has been merged you should be able to rebase your branch against master, update it here and the build should run again 💚

paulwaitehomeoffice:
On the Returns and Movements Planner project (RMP) we’ve had a need for this functionality too. We have a form where users can choose one of four values for time — anytime, before, after, and between — and then enter time values.

Because “between” essentially means “after” and “before”, we wanted to show both of our existing after and before sections when the user clicked on “between”, as well as an “and” that links the two sections.

before

between

I’ve actually copied and edited the GOV.UK show-hide-content.js along similar lines to this — should have checked for a pull request first!

paulwaitehomeoffice:
Just for the record, our use case described above (https://github.com/alphagov/govuk_frontend_toolkit/pull/383#issuecomment-292235986) no longer exists, as we’ve redesigned.

pcraig3:
Closing this since it's a year later and our problem has been resolved.

PR #384, opened 2017-02-22
Enlarge mobile clickable areas on breadcrumbs and beta banners

alecgibson:
On mobile and tablet, the breadcrumb links and links inside the beta banner can be hard to click, because they are quite small.

This change increases their clickable area (whilst leaving their display the same).

# Phase banner

## Before

phase-banner-before

## After

phase-banner-after

# Breadcrumbs

## Before

breadcrumbs-before

## After

breadcrumbs-after

This Pull Request replaces the same proposed change on [`static`](https://github.com/alphagov/static/pull/906).

Trello: https://trello.com/c/kSRKSoI2/326-review-styles-on-tablet-and-mobile-viewports-on-the-new-navigation-pages

### DIFF
--- a/stylesheets/_touch-friendly-links.scss
+++ b/stylesheets/_touch-friendly-links.scss
@@ -1,0 +1,10 @@
+// This mixin will expand the clickable area for a link to make it easier to click on a touch-enabled
+// device.
+
+@mixin touch-friendly-links($padding: 3px) {
+  a {
+    padding: $padding;
+    margin: -1 * $padding;
+    outline-color: transparent;
+  }
+}
--- a/stylesheets/design-patterns/_alpha-beta.scss
+++ b/stylesheets/design-patterns/_alpha-beta.scss
@@ -2,6 +2,7 @@
 @import "../typography";
 @import "../shims";
 @import "../grid_layout";
+@import "../touch-friendly-links";
 
 // Phase banner usage:
 //
@@ -38,6 +39,8 @@
     display: table-cell;
     vertical-align: baseline;
   }
+
+  @include touch-friendly-links();
 }
 
 // Phase tag usage:
--- a/stylesheets/design-patterns/_breadcrumbs.scss
+++ b/stylesheets/design-patterns/_breadcrumbs.scss
@@ -2,6 +2,7 @@
 @import "../typography";
 @import "../shims";
 @import "../url-helpers";
+@import "../touch-friendly-links";
 
 // Breadcrumbs usage:
 //
@@ -45,6 +46,8 @@
     }
 
   }
+
+  @include touch-friendly-links();
 
   a {
     color: $text-colour;


fofr:
Could you demonstrate the improvement with a mobile screenshot that shows the bounding area for the elements, before and after?

Seems like a nice fix 👍

alecgibson:
@fofr done. Some things to think about:

0. Do we need to worry about the fact that the focus state covers some text up now? (I assume we don't actually see this on mobile/tablet apart from onclick?)
0. Do we need to worry about multiple links in the phase banner overlapping their clickable areas? Since links further on in the DOM adopt a higher stacking order than previous links, it's conceivable that you could mostly overlap an earlier link?

nickcolley:
It'd be good to consider if this should be for under desktop only, touch interfaces are now available for not just small screens so this may make sense to increase the padding across the board, it could also make these easier to click with a cursor.

alecgibson:
@nickcolley I'd be fine with that, although if we do it on desktop too there's a higher chance that people will tab through links, which will enable this `focus` state (and could hide text)

nickcolley:
Since outlines do not count as a clickable area, we could remove them since we're replacing that with padding?

alecgibson:
I'm not sure I understand?

alecgibson:
So I've removed the outlines, but still unsure whether we should be including desktop breakpoints in this, as the larger `focus` state can cover some of the other text - this is especially noticeable when tabbing through links.

nickcolley:
I'm uncomfortable with making assumptions about people using touch based on viewport width but I'm not sure what's a better proposal at the moment so wont block this. 👍

alecgibson:
@nickcolley here's the focus states with no outline (so the orange area should now exactly correspond to the clickable area):

phase-banner-after-no-outline

breadcrumbs-after-no-outline

alecgibson:
@nickcolley also another argument for _not_ making assumptions based on viewport, is that I'm pretty sure that iPads in landscape will fall under the `desktop` breakpoint. I'd like a universal solution, but I'm not sure what that looks like - possibly having a separate focus styling?

The question of overlapping links in the Beta banner is still open, although maybe it's not worth worrying about until we encounter it in the wild with multiple links?

nickcolley:
One way I've tackled this in the past is to have an invisible pseudo element, this would not change the visual size of the elements but can be brittle if done badly.

fofr:
For future reference, CSS4 has: https://drafts.csswg.org/mediaqueries/#pointer
`@media (pointer:coarse) {`

alecgibson:
@nickcolley I've pushed a solution using pseudoelements, which keeps the original outline when tabbing, but has a larger clickable area across all breakpoints. I'm still concerned with links being above each other on subsequent lines, so I've reduced the effective area a little bit. I believe we have plans to remove all but one breadcrumb on mobile anyway, but I can't seem to find the Trello card.

alecgibson:
@nickcolley are you happy for me to merge this?

nickcolley:
I feel like if we're not confident about overlapping we should make sure this is a robust as possible, and maybe make a compromise. Since this is an invisible change I don't think it'd be very easy for us to know if that impacts people negatively.

Sorry for all the back and forth.

alecgibson:
@nickcolley what if - for now - we set the clickable area to match the outline (`3px`) as a compromise? That way there's no more overlapping than we already get with the outline, and the clickable area is more consistent with the visual cue?

alecgibson:
@nickcolley as discussed, I've gone back to the padding approach. Good to merge?

stylesheets/_touch-friendly-links.scss
```diff
@@ -0,0 +1,10 @@
+// This mixin will expand the clickable area for a link on mobile and tablet to make it easier to click on a touch
```
36degrees: Possibly a bit pedantic, but I can't see anything that limits this behaviour to only 'on mobile and tablet'? May be worth rewording to remove that clause.

stylesheets/_touch-friendly-links.scss
```diff
@@ -0,0 +1,10 @@
...
+// device.
+
+@mixin touch-friendly-links($padding: 3px) {
+  a {
+    padding: $padding;
+    margin: -1 * $padding;
+    outline: none;
```
36degrees: Does removing the outline cause any issues identifying the focussed element when the user has overridden text / background colours in their browser?

stylesheets/_touch-friendly-links.scss
alecgibson: Good spot - this got reworked to include desktop.

stylesheets/_touch-friendly-links.scss
alecgibson: Potentially - what's the use case here? A specialist browser? What can I test this in?

nickcolley: Setting the `outline-color` to `transparent` could avoid any potential issue with that.

alecgibson: I've just tried @nickcolley 's suggestion. Looks fine in my browser, but again I don't really know what the use case is here, or how to test it.

36degrees: The easiest way to test is in Firefox – https://support.mozilla.org/t5/Manage-preferences-and-add-ons/Change-the-fonts-and-colors-websites-use/ta-p/579

alecgibson: Okay so it can be done - but _why_ would people do it? If they're visually impaired and want to crank contrast? Or should that have been thought about in design already?

alecgibson: Tested on Firefox - looks fine on the default high contrast settings. The outline is put back in blue (I guess they override my override).

dsingleton:
I've not read the full background of the change, just skimmed, but I have a couple of comments:

1. Are we confident rolling this out through FET, and to any service using the breadcrumbs pattern, or would we want to trial this on GOV.UK Publishing first? 

  Traditionally code in FET has been battle tested somewhere else first - established patterns being uplifted. This _seems_ a little speculative, in that we think it'll be an improvement, but haven't tested it in the wild yet - is that right?

2. Is `touch-friendly-links` now part of the external API of FET? Is it intended for internal code reuse, or are we expecting folks consuming FET as a dependency to use it? Either way the docs should be more explicit, adding it to the [public docs](https://github.com/alphagov/govuk_frontend_toolkit/blob/master/docs/mixins.md) or marking as internal.

Overlapping seems to have been resolved, but if we went back to a larger click area that would be much more likely to affect breadcrumb - on pages with many/long [breadcrumbs that wrap on to multiple lines](https://www.gov.uk/get-court-funds-money-when-you-turn-18):

image

alecgibson:
@dsingleton 

1. I [originally opened this on `static`](https://github.com/alphagov/static/pull/906) and would be happy to put it back there (or even just in `collections`?!)

1. Again I have no strong opinions on available API, and haven't done much with this - happy for someone to make a call on this.

1. Breadcrumbs are slated to be [reduced to a single entry on mobile](https://github.com/alphagov/govuk-content-navigation/pull/122), which would eliminate breadcrumb overlapping problems.

edwardhorsford:
Suggest talking with @selfthinker who has been doing a lot of investigation recently in to changing colours / user stylesheets. AFAIK there are a few variations to test.

@alecgibson for context, *some* users may need to adjust colours to help improve readability. This is often related to dyslexia, but not all users who do this are dyslexic. In many cases it will be to *reduce* contrast or invert colours. Some users will have particular colours that work better for them. See [this blog post](https://accessibility.blog.gov.uk/2016/05/26/accessibility-and-me-marian-foley/) for an example.

stylesheets/_touch-friendly-links.scss
36degrees: Users with impaired vision might want to increase contrast (in addition to perhaps zooming in, or overriding the font) – conversely, someone with dyslexia might choose to _reduce_ the contrast by using e.g. a peach background which reduces the [visual stress](http://www.bdadyslexia.org.uk/dyslexic/eyes-and-dyslexia#Visual Stress).

dsingleton:
> I originally opened this on static and would be happy to put it back there (or even just in collections?!)

Sorry, didn't realise it started over there. Up to you, but if you have concerns i'd recommend testing them out on GOV.UK Publishing before pushing more widely (i'd say probably `static`, as you'd want consistent behaviour between GOV.UK Publishing apps).

> Breadcrumbs are slated to be reduced to a single entry on mobile, which would eliminate breadcrumb overlapping problems.

Nice! So once that hits GOV.UK Publishing and has been proven, it'd move up to the breadcrumb mixin here?

alecgibson:
> Nice! So once that hits GOV.UK Publishing and has been proven, it'd move up to the breadcrumb mixin here?

Yeah, I'd guess so.

So am I closing this and moving it _back_ to `static`?

dsingleton:
> So am I closing this and moving it back to static?

Thats up to you (sorry!). It's really down to how confident the team is in this being a reliable improvement / low risk of causing problems or needing to be changed based on real user feedback. I don't have enough context to make that decision :)

alecgibson:
@dsingleton given the amount of conversation that's arisen out of this change, it's probably best to be conservative; I'll move it to `static`.

alecgibson:
Moving this to `static`: https://github.com/alphagov/static/pull/911

PR #387, opened 2017-02-28
Add GOVUKTracker module

boffbowsh:
This has a similar API to the GA tracker module so it’s compatible with
the `GOVUK.Analytics` module. It sends data to the server by adding an
`img` tag for an empty gif with the payload data encoded in the
querystring of the `src`. This data will be picked up from the CDN logs.

It adds some additional data that the main GA tracker doesn’t add, such
as performance information and the GA client ID. This is because it’s
assumed that GA sends similar things by default anyway.

This data will be used alongside regular exports of content metadata to
provide performance analysts with more reliable content-focused
analytics data.

This is for review only at the moment to check my approach is reasonable.

TODO:
- [x] Add specs


javascripts/govuk/analytics/govuk-tracker.js
```diff
@@ -0,0 +1,115 @@
...
+        }
+      }
+    }
+
+    return $.param(data)
+  }
+
+  GOVUKTracker.prototype.sendToTracker = function (type, payload) {
+    var doc = global.document
+    $(doc).on('readystatechange', function () {
```
nickcolley: This looks odd, can you explain this approach?

nickcolley:
What has prompted this work? I'm not sure what problem is trying to be solved by adding this.

boffbowsh:
This is just a spike for now to capture some data. GA isn't useful enough for performance analysts to effectively measure the success of Q1, especially where content is concerned. We need to be able to match page views to actual content items, and also have the raw performance data, not sampled. Proposal https://docs.google.com/document/d/1Tw1gG_M08jAj1lXdq1vzdozxXGvbTU_SOjGXnxFAhcs/edit#

nickcolley:
Looks like our other trackers have specific functions, so for this I wonder if we should break this up to have a.

'real-user-measurement-tracker' etc.

javascripts/govuk/analytics/govuk-tracker.js
boffbowsh: We need to wait until the `readyState` is `complete` so we can get more timings, e.g. `domInteractive` and `domComplete`

boffbowsh:
The other modules are plugins, whereas this is a Tracker that is compatible with the `GOVUK.GoogleAnalyticsUniversalTracker` api, so can be used with `GOVUK.Analytics`. So I'm not sure it's the same thing.

javascripts/govuk/analytics/govuk-tracker.js
```diff
@@ -0,0 +1,115 @@
...
+  'use strict'
+
+  var $ = global.jQuery
+  var GOVUK = global.GOVUK || {}
+
+  var GOVUKTracker = function (gifUrl, fieldsObject) {
+    this.gifUrl = gifUrl
+    this.dimensions = []
+    global.ga(function (tracker) {
+      this.gaClientId = tracker.get('clientId')
```
fofr: What happens if GA isn't loaded?

boffbowsh: Good call, I'll add a guard

javascripts/govuk/analytics/govuk-tracker.js
```diff
@@ -0,0 +1,115 @@
...
+        if (typeof v === 'string' || typeof v === 'number') {
+          data['timing_' + k] = v.toString()
+        }
+      }
+    }
+
+    return $.param(data)
+  }
+
+  GOVUKTracker.prototype.sendToTracker = function (type, payload) {
```
fofr: Will this append an image for every event/pageview/tracker action?

fofr:
I wonder how the image loading would affect users leaving the page, eg clicking external links or download links that get tracked as events.

javascripts/govuk/analytics/govuk-tracker.js
fofr: Calling an asynchronous method from a repeatedly called synchronous one using local variables worries me. eg Value of local variable in asynchronous call may not be the one you expect. Not sure it's an issue here, specs should clarify what's happening.

fofr:
Presumably we are using a GIF rather than https://developer.mozilla.org/en-US/docs/Web/API/Navigator/sendBeacon because of current browser support?

javascripts/govuk/analytics/govuk-tracker.js
boffbowsh: Yeah. I wasn't sure whether to use an image tag or just simply AJAX the thing.

boffbowsh:
We could try `sendBeacon` and fall back to the `img` tag if the browser doesn't support it? But also, this is just meant to be a spike to get some representative data. It doesn't yet need to be 100% complete in this iteration.

javascripts/govuk/analytics/govuk-tracker.js
```diff
@@ -0,0 +1,117 @@
...
+    var doc = global.document
+    $(doc).on('readystatechange', function () {
+      if (doc.readyState === 'complete') {
```
tomnatt: As far as I remember once the page state reaches complete it doesn't change again. This restriction combined with responding only to changes in ready state means the tracker is going to miss any events or other actions triggered by a user within the page.

javascripts/govuk/analytics/govuk-tracker.js
tomnatt: Unless you're going to include a noscript option I'd have thought AJAX is a simpler approach

javascripts/govuk/analytics/govuk-tracker.js
tomnatt: Tests should clarify this

boffbowsh:
Specs added, ready for another review. Still todo, add the screen resolutions

boffbowsh:
Screen measurements added

javascripts/govuk/analytics/analytics.js
```diff
@@ -13,6 +13,11 @@
       delete config.universalId
       this.trackers.push(new GOVUK.GoogleAnalyticsUniversalTracker(universalId, config))
     }
+    if (typeof config.govukTrackerGifUrl !== 'undefined') {
```
tomnatt: I think this redundant since you switched to using AJAX in the `sendData` method.

boffbowsh: If the calling application doesn't configure govukTrackerGifUrl, this tracker shouldn't be added to the stack

fofr:
@boffbowsh Could you include a test that makes multiple calls?
Do you have an example of the network calls this makes when run on a page?

boffbowsh:
I think I'll have to change `sendBeacon` to a `$.get` only - nginx won't support `POST` to the `empty_gif` endpoint.

boffbowsh:
This is ready to go again now. I'll get it on integration again for final testing. Here's a HAR file as an example, viewable with http://www.softwareishard.com/har/viewer/.
[www.dev.gov.uk.har.zip](https://github.com/alphagov/govuk_frontend_toolkit/files/842152/www.dev.gov.uk.har.zip)

nickcolley:
Related: https://github.com/addyosmani/timing.js/

boffbowsh:
I'd like to get this live to start collecting some data, and test its usefulness. Can I get another review please @tomnatt / @nickcolley ?

javascripts/govuk/analytics/govuk-tracker.js
```diff
@@ -0,0 +1,128 @@
+;(function (global) {
+  'use strict'
+
+  var $ = global.jQuery
+  var GOVUK = global.GOVUK || {}
+
+  var GOVUKTracker = function (gifUrl, fieldsObject) {
```
tomnatt: `fieldsObject` doesn't get used anywhere. If for the first pass of this we're not passing it through, should chop it out entirely.

spec/unit/analytics/govuk-tracker.spec.js
```diff
@@ -0,0 +1,152 @@
...
+var $ = window.jQuery
+
+describe('GOVUK.GOVUKTracker', function () {
+  'use strict'
+  var GOVUK = window.GOVUK
+
+  var tracker
+
+  beforeEach(function () {
+    tracker = new GOVUK.GOVUKTracker('http://www.example.com/a.gif', {foo: 'bar'})
```
tomnatt: This constructor should only be taking a single argument.

tomnatt:
FWIW my analysis of the performance implications (throttled to a 3g connection):
* takes about 0.3s
* adds about 1/2kb
* runs async with (and within) other asset loading, so doesn’t actually extend page loading

Now awaiting a FEDEV to tell me I'm completely wrong 😃

boffbowsh:
@nickcolley can you comment/approve please?

nickcolley:
We're going to review this in three weeks as part of the two week blitz, for now we'll get this in so we can test the kind of data we're getting back.

In the future I think this probably should belong in Static and we should address the issues with cross browser APIs

PR #389, opened 2017-03-16
Deprecate selection-buttons.js

robinwhittleton:
GOV.UK Elements 3.0.0 will remove the need for GOVUK.SelectionButtons as it’s moving to a pure CSS solution. This adds documentation warnings that this is now deprecated and will be removed in future.

Probably worth waiting until 3.0.0 is released before merging this.

### DIFF
--- a/javascripts/govuk/selection-buttons.js
+++ b/javascripts/govuk/selection-buttons.js
@@ -1,3 +1,6 @@
+// DEPRECATED
+// This isn’t needed if you’re using GOV.UK Elements 3.0.0 or above
+
 ;(function (global) {
   'use strict'
 
--- a/javascripts/govuk/selection-buttons.js
+++ b/javascripts/govuk/selection-buttons.js
@@ -4,6 +4,10 @@
 ;(function (global) {
   'use strict'
 
+  if (window.console && window.console.warn) {
+    window.console.warn('Deprecation warning: Custom radio buttons and checkboxes (released in GOV.UK Elements 3.0.0) no longer require this JavaScript.')
+  }
+
   var $ = global.jQuery
   var GOVUK = global.GOVUK || {}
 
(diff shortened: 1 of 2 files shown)


javascripts/govuk/selection-buttons.js
```diff
@@ -1,3 +1,6 @@
+// DEPRECATED
+// This isn’t needed if you’re using GOV.UK Elements 3.0.0 or above
+
 ;(function (global) {
```
nickcolley: Could we add a `console.warn('Deprecated this is going away blah blah pls do the cool new way')`?

gemmaleigh:
screen shot 2017-03-17 at 12 12 24

javascripts/govuk/selection-buttons.js
```diff
@@ -1,6 +1,11 @@
...
   'use strict'
 
+  console.warn('Deprecation warning: Custom radio buttons and checkboxes (released in GOV.UK Elements 3.0.0) no longer require this JavaScript.')
```
nickcolley: Think we need to guard against console not being there since older browsers will blow up

http://caniuse.com/#search=console

gemmaleigh: @nickcolley is it sufficient to use `if (window.console) { ... }`?

nickcolley: We'd have to do the full whammy `if (window.console && window.console.warn) {}` like https://github.com/alphagov/govuk_prototype_kit/blob/6a38393eb657394fd21f7cbad2efb8ea8d709840/app/assets/javascripts/application.js#L7

gemmaleigh:
@nickcolley it would be good to merge this, now that GOV.UK elements v3.0.0 has been released and the custom radio buttons and checkboxes no longer require `selection-buttons.js`.

javascripts/govuk/selection-buttons.js
gemmaleigh: Hi @nickcolley, this has been added (see below).

nickcolley: Nice 👍

PR #390, opened 2017-03-17
show-hide-content should work with new radio/checkbox markup

robinwhittleton:
- update the selectors in show-hide-content to work with both the new and original markup. Simply removing the `.block-label` left it feeling a bit unscoped, so I’ve added a direct child selector. This potentially could cause problems, but it works with our recommended markup old and new at least.
- update the spec to follow the new markup.
- update the documentation to recommend the new markup.

Things to consider:
- are the new selectors too restrictive?
- should we be testing both old and new markup?
- how could we decouple this in future further?

### DIFF
--- a/docs/javascript.md
+++ b/docs/javascript.md
@@ -464,13 +464,15 @@ GOVUK.shimLinksWithButtonRole.init({
 
 Script to support show/hide content, toggled by radio buttons and checkboxes. This allows for progressive disclosure of question and answer forms based on selected values:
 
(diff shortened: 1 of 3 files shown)


gemmaleigh:
I'm keen to merge this as the examples of 'Conditionally revealing content' are currently broken for [GOV.UK elements](https://govuk-elements.herokuapp.com/form-elements/#form-toggle-content).

Let's get this in and bump the toolkit, if there are suggestions for a better way of achieving this, we can roll those out later.

PR #391, opened 2017-03-17
v5.1.2

gemmaleigh:
Update show-hide-content.js to work with new `.multiple-choice` custom
radio buttons and checkboxes, released in GOV.UK elements 3.0.0.


robinwhittleton:
We should probably call out https://github.com/alphagov/govuk_frontend_toolkit/pull/377 in the changelog too.

PR #393, opened 2017-03-17
Fixes for Travis when deploying

gemmaleigh:
This PR moves the open_ssl command back to before_install, otherwise it runs
more than once (as there is more than one deploy provider).

Related Travis issue here:
https://github.com/travis-ci/travis-ci/issues/2570

This behaviour was reverted in GOV.UK elements here:
https://github.com/alphagov/govuk_elements/pull/415

Before deploying, also check that the build has passed successfully.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,15 +5,15 @@ node_js:
 before_install:
   # Node project, so Ruby dependencies must be installed manually (see `govuk-lint`)
   - bundle install
-script:
-  - npm test
-before_deploy:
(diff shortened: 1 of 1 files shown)


dsingleton:
👍  LGTM. 

The `before_deploy` behaviour being per-deploy is somewhat unexpected :( 

We're going to be looking at https://github.com/alphagov/govuk_template/pull/267 with @alexmuller on Monday, lets see if we can find a nicer way for the future then.

PR #396, opened 2017-03-20
Release 5.1.3

boffbowsh:
Add an experimental GOV.UK analytics tracker. Given the URL to a single pixel GIF, it will encode analytics data as query string parameters on the call to the GIF.

@nickcolley


tomnatt:
👍

PR #401, opened 2017-04-13
Disable google analytics ad tracking

nickcolley:
I'm finding this difficult to test but this does seem to stop requests
to ga-audience and doubleclick.

Discovered when checking some waterfall logs, seen in external tools such as
SpeedCurve, WebPageTest.

To test I tried this locally, it seems these endpoints are only called on initial visit
which makes it hard to reproduce.

Fixes: https://github.com/alphagov/static/issues/991

Will need sign off from the Performance Analysis community

It'd be good to get someone else to try and reproduce this locally too.


mcgoooo:
🚢

Issue #402, opened 2017-05-02
show-hide-content throws an error for radio buttons with a period in their name attribute


pauldwaite:
### Steps to reproduce

1. Use the show-hide-content.js script with a radio button that has a period in its name attribute, for example:

   <label class="block-label selection-button-radio" for="choice_yes" data-target="choice_description_panel">
        <input id="choice_yes" type="radio" name="choice.yesno" value="yes">
        <span class="text">Yes</span>
    </label>

   <label class="block-label selection-button-radio" for="choice_no">
        <input id="choice_no" type="radio" name="choice.yesno" value="no">
        <span class="text">Yes</span>
    </label>

    <div class="panel panel-border-narrow js-hidden" id="choice_description_panel" aria-hidden="true">
        <label class="form-label" for="choice_description">Specify a meal type</label>
        <input type="text" class="form-control" id="choice_description" name="choice.description" value="">
    </div>

2. Click on the radio button that shows the content.

### Expected behaviour

The content is shown.

### Actual behaviour

A JavaScript error is thrown:

    Syntax error, unrecognized expression: [data-target] > input[type="radio"][name=choice.yesno][aria-controls]    jquery-1.11.3.js:1497

### Notes

The Scala Play framework’s form mapping feature allows form fields to be mapped to object properties, by putting periods in form field name attributes: https://www.playframework.com/documentation/2.5.x/ScalaForms#Nested-values

As such, it is common for Play web projects to have form fields with periods in their name attributes.

The HTML spec appears to allow name attributes containing periods: https://www.w3.org/TR/html5/forms.html#naming-form-controls:-the-name-attribute

36degrees:
I believe this was addressed by #403 (just need to ship a release to actually get the change out)

PR #403, opened 2017-05-02
Add support to show-hide-content.js for form fields with more varied …

paulwaitehomeoffice:
…characters (including periods) in their name attribute.

I’m not sure I added the tests in quite the right way — any advice welcome.

Fixes issue #402 https://github.com/alphagov/govuk_frontend_toolkit/issues/402

### DIFF
--- a/javascripts/govuk/show-hide-content.js
+++ b/javascripts/govuk/show-hide-content.js
@@ -14,12 +14,6 @@
       checkbox: '[data-target] > input[type="checkbox"]'
     }
 
-    // Escape name attribute for use in DOM selector
-    function escapeElementName (str) {
-      var result = str.replace('[', '\\[').replace(']', '\\]')
-      return result
-    }
-
     // Adds ARIA attributes to control + associated content
     function initToggledContent () {
       var $control = $(this)
@@ -79,7 +73,7 @@
     // Handle radio show/hide
     function handleRadioContent ($control, $content) {
       // All radios in this group which control content
-      var selector = selectors.radio + '[name=' + escapeElementName($control.attr('name')) + '][aria-controls]'
+      var selector = selectors.radio + '[name="' + $control.attr('name') + '"][aria-controls]'
       var $form = $control.closest('form')
       var $radios = $form.length ? $form.find(selector) : $(selector)
 
--- a/spec/unit/show-hide-content.spec.js
+++ b/spec/unit/show-hide-content.spec.js
@@ -32,6 +32,18 @@ describe('show-hide-content', function () {
         '<div id="show-hide-radios" class="panel js-hidden" />' +
         '</form>' +
 
+        '<form>' +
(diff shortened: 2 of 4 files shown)


spec/unit/show-hide-content.spec.js
```diff
@@ -30,6 +30,17 @@ describe('show-hide-content', function () {
         '<label>No</label>' +
         '</div>' +
         '<div id="show-hide-radios" class="panel js-hidden" />' +
+
```
romidane: Is there a reason why ever tag needs concatenating?

javascripts/govuk/show-hide-content.js
```diff
@@ -16,7 +16,7 @@
 
     // Escape name attribute for use in DOM selector
     function escapeElementName (str) {
-      var result = str.replace('[', '\\[').replace(']', '\\]')
+      var result = '"' + str.replace('"', '\\"') + '"'
```
romidane: This could just be 
```
function escapeElementName (str) {
  return '"' + str.replace('"', '\\"') + '"'
}
```

spec/unit/show-hide-content.spec.js
```diff
@@ -225,6 +273,10 @@ describe('show-hide-content', function () {
         }))
         expect(events && events.click).toContain(jasmine.objectContaining({
           namespace: 'ShowHideContent',
+          selector: 'input[type="radio"][name="test.single"]'
+        }))
+        expect(events && events.click).toContain(jasmine.objectContaining({
```
romidane: Could you split this test out?.. You seem to making and boolean evaluation inside an assertion?
Perhaps having a test which check that the event doesn't exist?

spec/unit/show-hide-content.spec.js
paulwaitehomeoffice: I presume it’s for readability — in the absence of multi-line strings, it’s a way to avoid all the HTML being on one line.

spec/unit/show-hide-content.spec.js
paulwaitehomeoffice: Yeah, could be. As I was only looking to fix a specific bug with this change, I wasn’t looking to redesign the tests as part of the same change.

romidane: That's fair enough. Seems like all the other tests do that too. Would probably better to refactor this separately you're right.

adamliptrot-oc:
Could this issue be fixed by just quoting the `name` attribute reference?
`var selector = selectors.radio + '[name="' + escapeElementName($control.attr('name')) + '"][aria-controls]'`

javascripts/govuk/show-hide-content.js
joelanman: might be wrong but won't this just replace one instance? might you need

`str.replace(/"/g, '\\"')`

instead?

PR #405, opened 2017-05-03
Use the more sensibly named govukTrackerUrl config name

boffbowsh:
Use `govukTrackerUrl` for configuration instead, as gifs aren’t
involved anywhere.


36degrees:
Does this need a corresponding change [in static](https://github.com/alphagov/static/blob/fc0b3ef10a246089562f6fe35aa8e6004b4d072f/app/assets/javascripts/analytics/init.js.erb#L18-L19)?

boffbowsh:
Yep, once the gem is released

PR #409, opened 2017-05-10
Increase button focus specificity to avoid being overridden

nickcolley:
We've updated govuk_template link focus styles which now overrides our buttons

screen shot 2017-05-10 at 14 04 53

As seen in https://www.gov.uk/book-driving-test

Introduced here: https://github.com/alphagov/govuk_template/commit/79466a489c38b0b82f4cb95daa2baba41b375a81

This kind of bug will hopefully be much more obvious when we merge these two repos as part of the new GOV.UK Frontend project :)

### DIFF
--- a/stylesheets/design-patterns/_buttons.scss
+++ b/stylesheets/design-patterns/_buttons.scss
@@ -87,6 +87,7 @@
     color: $white;
 
     &:link,
+    &:link:focus,
     &:hover,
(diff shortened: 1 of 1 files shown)


nickcolley:
Happy with this as I've tested end to end with https://github.com/alphagov/static/pull/1035

PR #420, opened 2017-06-20
Remove multivariate test javascript

fofr:
GOV.UK now has an alternative AB testing strategy which is server side. A server side approach prevents flashes of content. Deleting this logic will reduce the risk of creating tests that might be inaccessible, such as this: https://github.com/alphagov/frontend/pull/1228#pullrequestreview-37664574

Even when no tests are running, GOV.UK includes this JS on every page:
https://github.com/alphagov/static/blob/master/app/assets/javascripts/application.js#L4

After discussion with @nickcolley, on the belief that this has mainly been used by GOV.UK, we think this library could be removed. It would be a breaking change with a major version bump. Apps that still require the tests could copy the javascript across.

cc @carvil @alecgibson


nickcolley:
cc @chao-xian @ikennaokpala

carvil:
This sounds good to me @fofr ! When merged please update the manual page on A/B tests here: https://docs.publishing.service.gov.uk/manual/ab-testing.html

ikennaokpala:
Let me check with Polly on the content team about when the AB test should end cc\ @fofr @nickcolley

fofr:
AB test documentation updated here: https://github.com/alphagov/govuk-developer-docs/pull/292

PR #421, opened 2017-06-26
Add support for GA Universal

gunjam:
Google Analytics Universal provides a function `ga` that sends all
communication to Google Analytics. The `sendGoogleAnalyticsEvent`
function now checks if the `ga` function exists and if it does, uses it.

Port of commit [6bb47fd](https://git.io/vQs2t) by @robyoung

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,15 @@
 # Unreleased
 
-- Fix: Add support to show-hide-content.js for form fields with more varied characters: ([PR 403](https://github.com/alphagov/govuk_frontend_toolkit/pull/403))
+- Breaking change: Add support for Google Analytics Universal in stageprompt.js
+
+  This is a breaking change since while we allow for the old behaviour, Google Analytics Universal is used by the default, which could change behaviour in applications.
+
+  Also ensures that events triggered via GA Universal do not impact bounce-rate calculations,
+  which mirrors the original implementation.
+
+  ([PR 421](https://github.com/alphagov/govuk_frontend_toolkit/pull/421)).
+
+- Fix: Add support to show-hide-content.js for form fields with more varied characters: ([PR 403](https://github.com/alphagov/govuk_frontend_toolkit/pull/403)).
 
 # 7.6.0
 
--- a/javascripts/stageprompt.js
+++ b/javascripts/stageprompt.js
@@ -62,7 +62,11 @@
   }())
 
   GOVUK.performance.sendGoogleAnalyticsEvent = function (category, event, label) {
-    global._gaq.push(['_trackEvent', category, event, label, undefined, true])
+    if (global.ga && typeof(global.ga) === 'function') {
+      global.ga('send', 'event', category, event, label)
+    } else {
+      global._gaq.push(['_trackEvent', category, event, label, undefined, true])
+    }
   }
 
   global.GOVUK = GOVUK
--- a/javascripts/stageprompt.js
+++ b/javascripts/stageprompt.js
@@ -62,7 +62,7 @@
   }())
 
   GOVUK.performance.sendGoogleAnalyticsEvent = function (category, event, label) {
-    if (global.ga && typeof(global.ga) === 'function') {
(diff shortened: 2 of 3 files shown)


nickcolley:
I noticed in the previous implementation there's some tests, is it possible to get this tested here?

gunjam:
I've added in the missing unit tests from the previous version.

Ta.

robyoung:
Wow! @gunjam I am super impressed. How did you find that commit? I'd forgetten I even worked on that project.

nickcolley:
This looks good to me, but I think will be a breaking change since it'll always favour the ga global (which should be fine but is a breaking behaviour) 👍

rpowis:
Any chance of getting this merged? Some of our services have updated to universal analytics and I'd love to get rid of our [fork](https://github.com/hmrc/stageprompt) of the old repo.

nickcolley:
Going to ping the Analytics community at GDS to see if they're OK with this change.

javascripts/stageprompt.js
```diff
@@ -62,7 +62,11 @@
   }())
 
   GOVUK.performance.sendGoogleAnalyticsEvent = function (category, event, label) {
-    global._gaq.push(['_trackEvent', category, event, label, undefined, true])
+    if (global.ga && typeof global.ga === 'function') {
+      global.ga('send', 'event', category, event, label)
```
36degrees: The sixth element in the `trackEvent` array means means that `opt_noninteraction` is `true`, which [means that the 'event hit will not be used in bounce-rate calculation'](https://developers.google.com/analytics/devguides/collection/gajs/eventTrackerGuide#setting-up-event-tracking).

Assuming we'd want the same behaviour with GA Universal, I think we'd want to make this:

```
global.ga('send', 'event', category, event, label, {
  nonInteraction: true
})
```

as per https://developers.google.com/analytics/devguides/collection/analyticsjs/events#non_interaction_events?

I'm not a performance analyst however… there may be other better ways of achieving the same thing without hardcoding it.

nickcolley:
This doesnt seem to be used much internally at GDS, but rather a library used for external services: https://github.com/search?p=4&q=org%3Aalphagov+stageprompt&type=Code

javascripts/stageprompt.js
nickcolley: It seems sensible to mirror them.

@rpowis would be good to get your thoughts on this.

rpowis: yeah good spot @36degrees. i think this got overlooked for whatever reason in our forked version. 

so as things stand, any page using universal tracking with an event on it has been adding to the bounce rate if that event hasn't been triggered? 😬 

similar to the `_gaq` call, we probably don't want to add to the bounce rate in the majority of cases so defaulting to `{ nonInteraction: true }` would make sense. but i think having the ability to opt-in would be good.

that said, i don't think any services here are getting that granular with their analytics so i'm not sure it would have to be done as part of this PR.

CHANGELOG.md
```diff
@@ -1,6 +1,13 @@
 # Unreleased
 
-- Fix: Add support to show-hide-content.js for form fields with more varied characters: ([PR 403](https://github.com/alphagov/govuk_frontend_toolkit/pull/403))
+- Breaking change: Add support for GA Universal in stageprompt.js
```
36degrees: Can we explain why this is a breaking change, and what users need to do?

CHANGELOG.md
```diff
@@ -1,6 +1,13 @@
...
+
+  Also ensures that events triggered via GA Universal do not impact bounce-rate calculations,
```
36degrees: Does this need mentioning, as it's not a change from anything that's been released?

nickcolley: Given we know of forks that do not include this, I think it's worth mentioning.

CHANGELOG.md
nickcolley: Good shout will update

nickcolley:
I have spoken to @gunjam and got their approval for the updates I have made.

36degrees:
:shipit:

PR #424, opened 2017-07-04
Add package-lock.json

frankieroberto:
This is an artefact of npm 5+, which is an equivalent of the `Gemfile.lock` in Ruby-land, in that it captures the exact versions (with a version number and a hash for integrity checking) of all
sub-dependencies.

It’s recommended that this is checked in to source control, so that you can be sure that the exact same dependencies are used by all users (e.g. developers, continuous integration servers, staging, production, etc).

See http://blog.npmjs.org/post/161081169345/v500 for details.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,1825 @@
+{
+  "name": "govuk_frontend_toolkit_test_runner",
+  "version": "0.1.0",
+  "lockfileVersion": 1,
+  "dependencies": {
+    "abbrev": {
+      "version": "1.1.0",
+      "resolved": "https://registry.npmjs.org/abbrev/-/abbrev-1.1.0.tgz",
+      "integrity": "sha1-0FVMIlZjbi9W58LlrRg/hZQo2B8=",
+      "dev": true
+    },
+    "acorn": {
+      "version": "5.0.3",
+      "resolved": "https://registry.npmjs.org/acorn/-/acorn-5.0.3.tgz",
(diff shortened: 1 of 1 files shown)


frankieroberto:
@tvararu I just noticed that https://github.com/alphagov/govuk_elements has a `yarn.lock`, suggesting that it’s using [yarn](https://yarnpkg.com/en/) rather than npm for dependency management. I don't think there’s much of a difference now (since npm 5+ added its own lockfile format with integrity hash), but perhaps it’s worth being consistent? I think Yarn is marginally faster still, but I guess npm has the advantage of being the default.

gemmaleigh:
Hi @frankieroberto @tvararu, govuk-elements is only using Yarn in attempt to speed up builds using Travis CI. We've been using a package-lock.json file for govuk-frontend (which will replace govuk-elements in the future). I'd be inclined to stick with npm, I'll raise an issue to replace the Yarn lockfile with npm's package-lock.json file for govuk-elements, I agree they should be consistent.

frankieroberto:
@gemmaleigh cool – I think you should see similar build speeds if you switch Travis to use npm5 instead of yarn.

I'm still getting my head around the dependencies between elements, toolkit and templates, which has confused me no end, so it's good to hear about `govuk-frontend`. Hopefully that'll help!

36degrees:
Following the launch of the [GOV.UK Design System](https://design-system.service.gov.uk/), GOV.UK Frontend Toolkit will now only get major bug fixes and security patches, so I'm going to close this.

PR #425, opened 2017-07-06
README Syntax fix

frankieroberto:


36degrees:
Thanks for this!

There appears to be two links – the kernel.org link which appears just to be the man page for git submodule, and the git-scm which is a little more informative.

I _think_ the git scm link is a little more useful as an introduction to submodules, so could we update the link to point there instead? I think that's what was intended in e95e2b1a0d5c6bf373d29ad3ad7b5e1d096e481e when this invalid syntax was introduced.

frankieroberto:
@36degrees yeah, I think you’re right, the git-scm one is better. I've pushed another commit. Let me know if you want to me squash them. (Not sure what your policy is on that).

36degrees:
Brill, thanks! 👍

PR #429, opened 2017-07-18
Fixes media player volume button overlap

vanitabarrett:
Fixes [#853](https://github.com/alphagov/static/issues/853). Volume up and down buttons in the media player were overlapping the current volume percentage on focus. This fix keeps the large button touch area, but adjust the focus state to be smaller.

**Before: volume button focus state overlaps the current volume percentage:**
screen shot 2017-07-18 at 14 36 38

**After: volume button touch area remains the same as previously, but focus state is smaller so as not to obscure current volume percentage:**
This shows the new focus state:
screen shot 2017-07-18 at 14 14 51

But we can see that the touch area remains the same:
screen shot 2017-07-18 at 14 14 35

This means the volume controls continue to work on small devices where all icons are more closely located together:
screen shot 2017-07-18 at 14 15 34

### DIFF
--- a/stylesheets/design-patterns/_media-player.scss
+++ b/stylesheets/design-patterns/_media-player.scss
@@ -110,7 +110,7 @@
       .mute {
         position: absolute;
         bottom: 0;
-        left: 120px;
+        right: 0;
         height: 40px;
         width: 50px;
         padding-top: 1px;
@@ -141,7 +141,6 @@
         bottom: 0;
         height: 40px;
         width: 60px;
-        padding-top: 2px;
         font-size: 24px;
         line-height: 40px;
         text-align: center;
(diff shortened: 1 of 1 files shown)


stylesheets/design-patterns/_media-player.scss
```diff
@@ -150,25 +149,48 @@
...
           color: $light-blue;
+          background-color: transparent;
+          outline: none;
+
+          &:before {
+            background-color: $focus-colour;
+          }
+        }
+
+        &:before {
```
fofr: Clever solution 💯 
How does this look in IE8?

vanitabarrett: @fofr looks like the player never worked in IE8... at least according to BrowserStack, all I get is JS errors on the [original link](http://govuk-component-guide.herokuapp.com/components/govspeak/fixtures/with_youtube_embed/preview). Looks fine visually (with this fix), but none of the functionality works.

PR #436, opened 2017-11-06
Add custom dimension to be joined on by GA upload

elenatanasoiu:
For: https://trello.com/c/SYgBUl4q/258-develop-an-upload-to-ga-of-link-checker-data

## WHAT

This change is meant to duplicate the url information in a custom dimension. The url is usually sent in the event action. However, we need this info to also be present in a custom dimension.

## WHY
The reason for this is that we are doing a join between the click hit data and the information we upload via a Google Analytics Custom Upload. The custom upload is called "External Link Status" and needs to join existing data strictly on a custom dimension.

### DIFF
--- a/javascripts/govuk/analytics/external-link-tracker.js
+++ b/javascripts/govuk/analytics/external-link-tracker.js
@@ -21,6 +21,15 @@
         options.label = linkText
       }
 
+      // This custom dimension will be used to duplicate the url information
+      // that we normally send in an "event action". This will be used to join
+      // up with a scheduled custom upload called "External Link Status".
+      // We can only join uploads on custom dimensions, not on `event actions`
+      // where we normally add the url info.
+      var googleAnalyticsExternalLinkUploadDimension = 36
+      var urlDimensionToJoinUploadOn = href
+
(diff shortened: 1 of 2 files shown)


javascripts/govuk/analytics/external-link-tracker.js
```diff
@@ -21,6 +21,15 @@
...
 
+      // This custom dimension will be used to duplicate the url information
+      // that we normally send in an "event action". This will be used to join
+      // up with a scheduled custom upload called "External Link Status".
+      // We can only join uploads on custom dimensions, not on `event actions`
+      // where we normally add the url info.
+      var googleAnalyticsExternalLinkUploadDimension = 36
+      var urlDimensionToJoinUploadOn = href
+
+      GOVUK.analytics.setDimension(googleAnalyticsExternalLinkUploadDimension, urlDimensionToJoinUploadOn)
```
fofr: It might be better to make the dimension an option that can be specified here:
https://github.com/alphagov/static/blob/master/app/assets/javascripts/analytics/static-analytics.js#L27

elenatanasoiu: I've created two new PRs for this here: 

https://github.com/alphagov/govuk_frontend_toolkit/pull/439
https://github.com/alphagov/static/pull/1192

PR #438, opened 2017-11-09
Version 7.2.0

elenatanasoiu:
Add custom dimension on `TrackEvent` to duplicate the url information that we normally send on a the `event action`. This will be used to join up with a scheduled custom upload called "External Link Status". We can only join uploads on custom dimensions, not on `event actions`, where we normally add the url info.

PR https://github.com/alphagov/govuk_frontend_toolkit/pull/436


36degrees:
Sorry, I totally should have picked up on this when #436 was opened – GOV.UK Frontend Toolkit is used by other departments within government. I don't _know_ if any of them are have adopted this code, but if they have, we'll be dumping data into one of their GA dimensions, whether they want it or not. Is there any way it could be adapted to be opt-in behaviour that GOV.UK can choose to use?

robinwhittleton:
I was never sure why GA was part of the toolkit itself and not a seperate repo…

fofr:
It was moved into the toolkit so it could be used across all GOV.UK apps, which at the time included Rails and Node.js apps – the toolkit was the obvious home without creating something new. It was also then used by Digital Marketplace.

It should be split out into it's own repo.

36degrees:
We currently have unanswered questions about what happens with these analytics scripts in a GOV.UK Frontend world, where teams are otherwise no longer using Frontend Toolkit.

It would be *awesome* if at least some of those questions went away in the next few months.

However, I worry we're going off on a tangent here that doesn't directly relate to this PR.

elenatanasoiu:
I'm going to open a new PR to make the custom dimension feature optional, and turned off by default. Then I can turn it on in static, similar to what was done here: https://github.com/alphagov/static/pull/1183/commits/b98fc5033567026e64f759d3559772ade056490e

elenatanasoiu:
I've created two news PRs here if you want to have a look: 

I've created two new PRs for this here:

https://github.com/alphagov/govuk_frontend_toolkit/pull/439
https://github.com/alphagov/static/pull/1192

elenatanasoiu:
The change to make the custom dimension optional has now been merged. You can see it here: https://github.com/alphagov/govuk_frontend_toolkit/pull/439

Which means this commit can now be used to bump the gem version. 🎉

PR #439, opened 2017-11-10
Allow custom dimension for uploads to be optional

elenatanasoiu:
For: https://trello.com/c/SYgBUl4q/258-develop-an-upload-to-ga-of-link-checker-data

Previous work (seen here: https://github.com/alphagov/govuk_frontend_toolkit/pull/436) has been done to allow us to duplicate the url information we normally store in an `event action`. We duplicate this info into a custom dimension in order to be able to do a join later within Google Analytics with a custom upload setup to report the status of links.
We weren't able to do the join on the info in an event action, only on a custom dimension.

However, it's been pointed out that the `govuk_frontend_toolkit` gem is being used for other projects besides `gov.uk` so would inadvertently use their custom dimension to duplicate this url info, without their approval.
So we've made this behavior optional.
If a custom dimension is not provided, then the behavior is turned off.

Depends on https://github.com/alphagov/static/pull/1192

### DIFF
--- a/javascripts/govuk/analytics/external-link-tracker.js
+++ b/javascripts/govuk/analytics/external-link-tracker.js
@@ -5,7 +5,9 @@
   var GOVUK = global.GOVUK || {}
 
   GOVUK.analyticsPlugins = GOVUK.analyticsPlugins || {}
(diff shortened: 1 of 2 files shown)


36degrees:
👌

PR #447, opened 2018-02-08
Add Django consumption docs

### LINKED ISSUE #442: Document recommended way to use this with python / django
There are a couple of packages on pypi

https://pypi.python.org/pypi/govuk-template

https://pypi.python.org/pypi/django-govuk-template/0.2

The second one of these says

```


It should be easy to make a Django-based service that follows Government Digital Services’ style guide and reference materials. But https://pypi.python.org/pypi/govuk-template is not kept updated and is not readily usable in Django and https://github.com/alphagov/govuk_template cannot be installed directly into a Django project without complex build steps.

This package takes components published by GDS and creates an app in a Django project which can then be used as normal. This process downloads a release from https://github.com/alphagov/govuk_template and the contents of Node.js packages https://www.npmjs.com/package/govuk-elements-sass and https://www.npmjs.com/package/govuk_frontend_toolkit
```

This project should document the recommended way of integrating with django, and ideally - as part of it's release process update the first one of those packages.

If you can't get hold of the maintainer of the pypi package, then it is worth opening a ticket here
https://github.com/pypa/pypi-legacy/issues

igloosi:
Fixes #442


README.md
```diff
@@ -39,6 +39,19 @@ may need to upgrade to a more recent version to use the grid helpers. Minimal
 compatible versions include `node-sass` 1.0.0, `grunt-sass` 0.16.0,
 `gulp-sass` 1.2.0 and `libsass` 3.0.0.
 
+### Django
+
+The easiest way to integrate it would be to create a `package.json` file in your application with `npm init` 
+You need to have [NodeJS](https://nodejs.org/en/) installed.
```
nickcolley: "Why do I need NodeJS? Does this give me NPM?"

igloosi:
@nickcolley updated

PR #459, opened 2018-04-20
Optionally strip dates from data sent to Google Analytics

barrucadu:
Some URLs can include dates of birth, which we should not sent to Google Analytics.  For example: https://www.gov.uk/state-pension-age/y/age/1991-09-09/male.  Previous work implemented optional postcode stripping, so I just copied that but for dates.  The regex matches YYYY-MM-DD style dates, with the dashes optional.

Same approach as #435

---

[Trello card](https://trello.com/c/zZzJHa8j/99-strip-dob-from-smart-answer-urls-before-sending-to-ga-2)


barrucadu:
Tim Leighton-Boyce from the performance analysis community spent Wednesday and Thursday last week on this, and is happy with it.

PR #462, opened 2018-06-11
Add notice message

igloosi:
The GOV.UK Design System is launching 22 June
===============

When it's launched the Design System will contain the tools and helpers for building frontend apps.

If you’re starting a new project and you’re able to wait until 22 June, we recommend using the Design System when it launches.

GOV.UK Frontend Toolkit will still be available in case you are already using it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,12 @@
+The GOV.UK Design System is launching 22 June
+===============
+
+When it's launched the Design System will contain the tools and helpers for building frontend apps.
+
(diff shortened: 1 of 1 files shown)


andysellick:
@igloosi can this change include a link to the Design system?

36degrees:
Whilst the Design System is still behind basic auth it probably doesn't make sense to – we are going to be updating this messaging in a couple of weeks once we're in 'public beta', so we'll add links then.

PR #466, opened 2018-07-16
Strip the user's email address out of the location used by GA

brucebolt:
We have seen some paths in Google Analytics that include the user's email address, normally as part of the query string.  These URLs always have zero pageviews, but a number of hits (GA measures hit as any activity, e.g. event, timing measurement, etc).

Although there is existing code in this repo to strip personal information (PII) from the path, there are some cases where GA generates a new request (e.g. page timings) that we have not covered.  In this case, it reports the actual URL rather than our custom path.  This aligns with the observation of these hits having no pageviews, since pageviews are already stripped of this information.

This PR addresses that problem by stripping email addresses out of all URLs and setting this as a global location for GA to use on all requests from that page.

https://trello.com/c/jKitCyZI/311-remove-emails-from-ga-events-on-email-alert-frontend-pages

### DIFF
--- a/javascripts/govuk/analytics/google-analytics-universal-tracker.js
+++ b/javascripts/govuk/analytics/google-analytics-universal-tracker.js
@@ -20,6 +20,10 @@
       sendToGa('set', 'displayFeaturesTask', null)
     }
 
+    function stripLocationPII () {
+      sendToGa('set', 'location', stripEmailAddressesFromString(window.location.href))
+    }
+
     // Support legacy cookieDomain param
     if (typeof fieldsObject === 'string') {
       fieldsObject = { cookieDomain: fieldsObject }
@@ -28,6 +32,7 @@
     configureProfile()
     anonymizeIp()
     disableAdTracking()
+    stripLocationPII()
   }
 
   GoogleAnalyticsUniversalTracker.load = function () {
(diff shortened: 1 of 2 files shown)


nickcolley:
Since this keeps coming up is there an opportunity to avoid this in the server side applications?

For example if we know certain routes allow user input:

```
/regular/route/:user-input
```

output this route to the frontend as a sort of 'mask' which then can take the current URL and remove this section from the URL.

`/regular/route/email@address.com` -> `/regular/route/[redacted]`

This means we're not relying on matching arbitrary patterns, which as we've seen seems prone to letting things slip through..

javascripts/govuk/analytics/google-analytics-universal-tracker.js
```diff
@@ -20,6 +22,10 @@
       sendToGa('set', 'displayFeaturesTask', null)
     }
 
+    function stripLocationPII () {
+      sendToGa('set', 'location', stripPIIFromString(window.location.href.toString()))
```
thomasleese: I think `window.location.href` is already a string so we don't need to call `toString`.

javascripts/govuk/analytics/google-analytics-universal-tracker.js
```diff
@@ -160,6 +167,11 @@
     }
   }
 
+  function stripPIIFromString (string) {
+    var stripped = string.replace(EMAIL_PATTERN, '[email]')
```
thomasleese: I wonder if the `EMAIL_PATTERN` variable should be in this function so it's closer to the code where it gets used and then maybe rename this to `stripEmailAddressesFromString`?

nickcolley:
@bevanloon sure, I won't block this pull request just wanted to highlight that previous to this we've already had two 'temporary band-aids', so we should probably try a more robust approach in the future.

javascripts/govuk/analytics/google-analytics-universal-tracker.js
brucebolt: Ah, of course.  I was thinking of `window.location`, which is an Object.

PR #476, opened 2019-04-23
Update MHLCG brand colour

edwardkerry:


36degrees:
Hey,

Just as a heads up, this repo’s been deprecated and is being replaced with GOV.UK Frontend.

It’d be great if you could [update the organisation colour in GOV.UK Frontend](https://github.com/alphagov/govuk-frontend/blob/master/src/settings/_colours-organisations.scss) as well so that we keep the two in sync.

Thanks!

PR #477, opened 2019-05-03
Disable Google Analytics Ad Features by default

elliotcm:
https://trello.com/c/D1PtNw0h/898-make-sure-we-cant-accidentally-send-extra-user-data-to-google


nickcolley:
Google Analytics adding more on by default features that compromise our users privacy is very frustrating :(

PR #484, opened 2020-03-06
🚧 Add tracking to see which JavaScript is being used and where

injms:
- Adding a Google Analytics event to see where, if anywhere, these JavaScript
   modules are being used.
 - Respects the user's settings by checking for the presences of the Google
   Analytics object - this won't be present if a user hasn't opted into
   tracking.

### DIFF
--- a/javascripts/govuk/details.polyfill.js
+++ b/javascripts/govuk/details.polyfill.js
@@ -6,7 +6,7 @@
 
 // http://www.sitepoint.com/fixing-the-details-element/
 
-;(function (global) {
+; (function (global) {
   'use strict'
 
   var GOVUK = global.GOVUK || {}
@@ -234,6 +234,13 @@
     init: function ($container) {
       GOVUK.details.addEvent(document, 'DOMContentLoaded', GOVUK.details.addDetailsPolyfill)
       GOVUK.details.addEvent(window, 'load', GOVUK.details.addDetailsPolyfill)
+
+      if (window.GOVUK.analytics && window.GOVUK.analytics.trackEvent) {
+        GOVUK.analytics.trackEvent('Toolkit', 'details.polyfill.js', {
+          label: window.location.pathname,
+          nonInteraction: true
+        })
+      }
     }
   }
   global.GOVUK = GOVUK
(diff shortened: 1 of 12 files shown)


injms:
Though this pull request is a draft, this is ready for review. It's marked as draft to prevent it being merged in as this is a temporary thing.

javascripts/govuk/details.polyfill.js
```diff
@@ -234,6 +234,15 @@
     init: function ($container) {
       GOVUK.details.addEvent(document, 'DOMContentLoaded', GOVUK.details.addDetailsPolyfill)
       GOVUK.details.addEvent(window, 'load', GOVUK.details.addDetailsPolyfill)
+
+      if (window.GOVUK.analytics && window.GOVUK.analytics.trackEvent) {
+        if (window.location.origin === 'https://gov.uk') {
```
andysellick: Just to check, because I don't know, should this be `https://www.gov.uk`?

Also wondering if this is only a temporary branch and will never be merged why we need this check?

javascripts/govuk_toolkit.js
```diff
@@ -1 +1,13 @@
 // = require_tree ./govuk
+
+(function () {
+  if (window.GOVUK.analytics && window.GOVUK.analytics.trackEvent) {
+    if (window.location.origin === 'https://gov.uk') {
+      GOVUK.analytics.trackEvent('Toolkit', 'stop-scrolling-at-footer.js', {
```
andysellick: I think this is a copy paste error from the previous file? `stop-scrolling-at-footer.js`?

javascripts/govuk/details.polyfill.js
injms: This has been removed.

javascripts/govuk_toolkit.js
injms: Fixed.

nickcolley:
> Respects the user's settings by checking for the presences of the Google
Analytics object - this won't be present if a user hasn't opted into
tracking.

You should consider a more explicit opt-in than checking to see if the analytics module is present.

injms:
This has served it purpose, so I can now close it.

